PR #336, opened 2013-10-15
Unecessary Whitespace

kpdecker:
I have come across an issue with the each block helper. It appears that it is creating unecessary whitespace. I copied the each block helper example (http://handlebarsjs.com/#iteration) exactly and my results were not the same. My results ended up like the following:

```
<ul>

  <li>ele1</li>

  <li>ele2</li>

  <li>ele3</li>

</ul>
```

instead of the expected:

```
<ul>
  <li>ele1</li>
  <li>ele2</li>
  <li>ele3</li>
</ul>
```

This might seem like a non issue but say we have a template like the following:

```
<script id="onlyEachTemplate" type="text/x-handlebars-template">
{{#each collection}}
<div>{{this}}</div>
{{/each}}
</script>
```

My expected output would be:

```
<div>ele1</div>
<div>ele2</div>
<div>ele</div>
```

but I instead get:

```
<div>ele1</div>

<div>ele2</div>

<div>ele3</div>
```

In an AJAX application where regular polling for data and rendering that data with a template like the previous can result in a large amount of unecessary dom growth. We actually encountered such an issue. All of the extra whitespace elements would spike the CPU usage under FF15 100% after 12 hours.
Here is a link to a set of tests I wrote to demonstrate the issue https://github.com/jayfunk/Handlebars-Whitespace-Test

### DIFF
--- a/spec/whitespace-control.js
+++ b/spec/whitespace-control.js
@@ -2,61 +2,61 @@ describe('whitespace control', function() {
   it('should strip whitespace around mustache calls', function() {
     var hash = {foo: 'bar<'};
 
-    shouldCompileTo(' {{(foo)}} ', hash, 'bar&lt;');
-    shouldCompileTo(' {{(foo}} ', hash, 'bar&lt; ');
-    shouldCompileTo(' {{foo)}} ', hash, ' bar&lt;');
+    shouldCompileTo(' {{~foo~}} ', hash, 'bar&lt;');
+    shouldCompileTo(' {{~foo}} ', hash, 'bar&lt; ');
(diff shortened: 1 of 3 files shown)


stan5079:
A bit late to the party but this might help some people.

I stumbled upon the same issue as @donaldpipowitch seems to be having. There were empty **lines** after certain handlebars. The issue for me was not the white**spaces** that this topic seems to be about.

In order to get rid of the empty lines, I opened the text block in the HTML editor and removed the `<br> `tags after the statements I don't want empty lines after. After saving, SendGrid will notify you that your text block will be changed into a code block.

Hope this helps.

PR #182, opened 2014-01-17
Context different for partial

kpdecker:
Hi folks,

When I have a template, say:

```
<div>
  {{#each updates}}
    <div class="update">
        {{#if ../user}}
          <div class="name">{{user.name}}</div>
        {{/if}}
        <p>{{text}}</p>
   </div>
  {{/each}}
</div>
```

And it works fine, but I move the update into  a partial:

**file: update.handlebars**

```
<div class="update">
  {{#if ../user}}
    <div class="name">{{user.name}}</div>
  {{/if}}
  <p>{{text}}</p>
</div>
```

And include the partial where the update code used to be:

```
<div>
  {{#each updates}}
    {{> update}}
  {{/each}}
</div>
```

I get "TypeError: Cannot read property 'user' of undefined" now. Why would the context be different for a partial containing the exact same code?

### DIFF
--- a/spec/partials.js
+++ b/spec/partials.js
@@ -24,10 +24,10 @@ describe('partials', function() {
   });
 
   it("partials with parameters", function() {
-    var string = "Dudes: {{#dudes}}{{> dude otherDude=name}}{{/dudes}}";
-    var partial = "{{otherDude}} ({{url}}) ";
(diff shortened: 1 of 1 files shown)


jfbrennan:
Where is this documented? If it's not, then I would gladly send the pull request for updating the docs if it meant one less developer had to go implement his own 'include' helper ;)

kpdecker:
@jfbrennan I just pushed docs for this to the site but glad to take any improvements you might be able to offer!

Issue #250, opened 2012-05-31
Include index in each block helper, aka each_with_index


rbu:
It would be really nice to have access to the index of an each iteration.
There are ready-to-use block helpers out there, but having this upstream would be nice.

I can provide a pull request if that helps.
## See also

https://gist.github.com/1048968
https://gist.github.com/1966803
http://stackoverflow.com/questions/5021495/in-mustache-how-to-get-the-index-of-the-current-section

aaronharding:
If it helps, this was my solution. I edited the execIteration function within the each helper.

handlebars-3.0.3, line 2375
```javascript
...
data.index = index;
data.humanIndex = index + 1; // added
data.first = index === 0;
...
```

This could very easily be added in with a pull request.

jblotus:
+1 for one based index

kpdecker:
Locking this thread as we are set on the implementation here. Zero-based index will continue to be exposed from the native implementation. Those that need a 1-based index should write a simple helper to add this value. With subexpressions that helper can be used as arguments to other helpers too so this provides the option to have one-based anywhere without the overhead of tracking another field in the iteration logic that may or may not be used or breaking existing templates.

```
Handlebars.registerHelper('addOne', function(value) {
  return value + 1;
});
```
```
Like This: {{addOne @index}} or {{helper (addOne @index)}}
```

PR #720, opened 2014-01-27
Add indexPlusOne support

cfjedimaster:


Ghazgkull:
Surprised and disappointed that Handlebars doesn't provide a one-based index out of the box. It's such a common and obvious need. Judging by the fact that #250 was locked, it looks like this issue is blocked by a personal desire to "win" the discussion. 👎

stevenvachon:
Surprised? If `@index` were 1 and not 0, `{{#if @index}}` would never be false. Have you used mustache?

cfjedimaster:
@stevenvachon to be clear - we're asking about support for a *new* var, indexPlusOne, or some such, that would not change the behavior of index.

stevenvachon:
@cfjedimaster my response was to @Ghazgkull -- I should've put that in.

cfjedimaster:
Fair enough.

Ghazgkull:
This discussion (and the others like it that have been locked) isn't about changing @index to be one-based. @cfjedimaster understood the context.

ErisDS:
Just to throw 2c in here - Ghost has its own version of each, [foreach](https://github.com/TryGhost/Ghost/blob/master/core/server/helpers/foreach.js#L76). Amongst other things it has this feature, and we called it `{{@number}}`.

PR #569, opened 2014-02-09
Unable to lookup array values using @index

kpdecker:
When we are working with {{#each}}  helper we are facing following issue.

JSON: {arr1 : ["a","v"],arr2 : ["c","d"]}

Template:
{{#each arr1}}
{{../arr2.[0]}}
{{/each}}

Above code is  working fine. But when we use @index insteated of static index number in the arr2 its not working. we couldn't understand why @index value is not placed. 

For Example (Not working),

Template:1
{{#each arr1}}
{{../arr2.[@index]}}
{{/each}}

Template:2
{{#each arr1}}
{{../arr2.[index]}}
{{/each}}

### DIFF
--- a/lib/handlebars/base.js
+++ b/lib/handlebars/base.js
@@ -193,6 +193,10 @@ function registerDefaultHelpers(instance) {
     var level = options.data && options.data.level != null ? parseInt(options.data.level, 10) : 1;
     instance.log(level, context);
   });
+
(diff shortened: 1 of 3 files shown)


snreynolds:
This took me forever to figure out.. but for anyone coming across this issue the syntax that worked for me was 
{{#lookup ../arr2 @index}}{{/lookup}}

PR #491, opened 2014-02-10
For nested helpers: get the @ variables of the outer helper from the inner one

kpdecker:
Following [this conversation](https://github.com/wycats/handlebars.js/issues/250#issuecomment-15836544), I think it would be very handy to be able, in nested helpers, to get the @ variables of the outer helper from the inner one.

```
{{#each outer}}
    {{#each inner}}
        <!-- How to get the outer index here? -->
    {{/each}}
{{/each}}
```

I am not sure what should be the syntax though...

```
../@index
@../index
```

Thanks

### DIFF
--- a/lib/handlebars/base.js
+++ b/lib/handlebars/base.js
@@ -223,5 +223,7 @@ export var logger = {
 export function log(level, obj) { logger.log(level, obj); }
 
 export var createFrame = function(object) {
-  return Utils.extend({}, object);
+  var frame = Utils.extend({}, object);
+  frame._parent = object;
+  return frame;
 };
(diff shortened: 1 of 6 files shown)


kuncevic:
any ideas how would you access the outer `@index` with https://github.com/webroo/dummy-json
like I have
```
{
"thing": [
  {{#repeat 10}}
  {
    "thingid": {{@index}}
    "otherThing": [
      {{#repeat 5}}
      {
      "thingId": {{@index}} ///need outer index here 'thingid'
      }
      {{/repeat}}
    ]
    }
  {{/repeat}}
}
```
the `@../index` does not work

PR #731, opened 2014-07-07
Strange behavior for {{#foo}} {{bar}} {{/foo}} when foo is 0

kpdecker:
```
var tmpl = Handlebars.compile('{{#foo}} This is {{bar}} ~ {{/foo}}');
alert(tmpl({ foo: 0, bar: 'OK' }));
```

in handebars.js 1.0.21 (or older version) we get output: ' This is  ~ '
in handlebars.js 1.1.0 (or newer version) we get output ' This is 0 ~ '

Is this a designed behavior or bug?
- When {{#foo}} section and foo === 0 , all {{any_name}} inside the section will be rendering as 0.

### DIFF
--- a/bench/templates/paths.js
+++ b/bench/templates/paths.js
@@ -1,7 +1,7 @@
 module.exports = {
-  context: { person: { name: "Larry", age: 45 } },
-  handlebars: "{{person.name}}{{person.age}}{{person.foo}}{{animal.age}}",
-  dust: "{person.name}{person.age}{person.foo}{animal.age}",
-  eco: "<%= @person.name %><%= @person.age %><%= @person.foo %><% if @animal: %><%= @animal.age %><% end %>",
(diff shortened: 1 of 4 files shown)


theodorejb:
There's a good case for treating `0` as non-empty for a section, since this is a visible value to render. However, treating a blank string as non-empty does not make sense, and causes a major compatibility issue with Mustache templates. E.g. a common pattern I've seen in Mustache templates is the following:

```hbs
{{#id}} id="{{ id }}"{{/id}}
```

Where the `id` attribute should only be rendered when it is a non-empty string. The current Handlebars.js behavior of rendering `id=""` in these cases breaks so many templates that it can block moving from Mustache to Handlebars.

I think this behavior should be changed - at a minimum when `compat` is enabled, but arguably always for consistency.

PR #892, opened 2014-10-28
Implement parser for else chaining of helpers

### LINKED ISSUE #72: feature request: elsif
of course it can be done with the existing helpers (if, else, unless), but often elsif is just the natural choice people are used too.

kpdecker:
Allows users to chain multiple helpers together using their inverse callbacks.
I.e.

```
{{#if foo}}
{{else if bar}}
{{else}}
{{/if}}
```

The control flow here effectively causes the helpers to be nested. The above
is actually syntactic sugar for this:

```
{{#if foo}}
{{else}}
 {{#if bar}}
 {{else}}
 {{/if}}
{{/if}}
```

Any helper may be used in this manner, the only requirement is they support
normal calls and inverse calls.

Introduces a breaking change in that `{{else foo}}` may no longer be used as a
root level operator. Instead `{{^foo}}` must be used.

Fixes #72.

### DIFF
--- a/lib/handlebars/compiler/helpers.js
+++ b/lib/handlebars/compiler/helpers.js
@@ -10,11 +10,29 @@ export function stripFlags(open, close) {
 
 export function prepareBlock(mustache, program, inverseAndProgram, close, inverted, locInfo) {
   /*jshint -W040 */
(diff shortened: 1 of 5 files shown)


kpdecker:
Released in 3.0.0

PR #896, opened 2014-11-02
Simplify BlockNode by removing intermediate MustacheNode

mmun:
This PR tweaks the AST for blocks. For example `{{#each items}}{{/each}}` is now parsed as

```
- block:
  - sexpr: (each items)
```

instead of

```
- block:
  - mustache:
    - sexpr: (each items)
```


kpdecker:
Released in 3.0.0

PR #902, opened 2014-11-08
Generate Source Maps

kpdecker:
Allows the precompiler to optionally generate source maps for the templates.

This is a breaking change for any users who are tightly integrated with the JavascriptCompiler class as the intermediate code format has been changed to use an object tree rather than outright string concatenation.


kpdecker:
Released in 3.0.0

PR #906, opened 2014-11-12
Add parser support for block params

mmun:
Adds basic support for block params on blocks:

```
{{#block ... as |foo|}}...{{/block}}
{{^block ... as |foo|}}...{{/block}}
{{#block ... as |foo bar|}}...{{/block}}
{{^block ... as |foo bar|}}...{{/block}}
```

The block params are stored as an array of strings on the program node.

For example `{{#block ... as |foo bar|}}...{{/block}}` generates a program node with a `blockParams` property set to `[ 'foo', 'bar' ]`.

### DIFF
--- a/lib/handlebars/compiler/ast.js
+++ b/lib/handlebars/compiler/ast.js
@@ -9,10 +9,11 @@ function LocationInfo(locInfo) {
 }
 
 var AST = {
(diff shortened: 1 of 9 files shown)


kpdecker:
Released in 3.0.0

PR #918, opened 2014-12-07
Add instructions for 'spec/mustache' to CONTRIBUTING.md, fix a few typos

oneeman:
I was going to just fix a few typos, but I ran into trouble running `grunt` due to not having anything in `spec/mustache`.  So I also added instructions on that to `CONTRIBUTING.md`.  I figured it's fine to do one PR rather than two since the typo fixes are trivial.


kpdecker:
Released in 3.0.0

PR #930, opened 2014-12-26
Add parent tracking and mutation to AST visitors

### LINKED ISSUE #916: Expand the Visitor API
We should allow for context so the node is aware of what the parent is and also allow for some mechanism for replacement of the currently visited node.

kpdecker:
Fixes #916


kpdecker:
Released in 3.0.0

Issue #935, opened 2015-01-06
A different Handlebars.print for re-printing an AST back to source code


fivetanley:
I'm not quite sure what to call it yet, other than Handlebars.print. Currently, Handlebars.print prints stuff like "BLOCK PARAMS[actual code]" https://github.com/wycats/handlebars.js/blob/master/lib/handlebars/compiler/printer.js#L30

It'd be great to have a utility function for just reprinting an AST. The reason for this being I'm working on some tools to help users of Ember upgrade some of their deprecated syntax and check the result of that back into version control without otherwise changing the source code. Ideally, you could parse a template, pass the AST into this function, and have the string returned be identical to the template you parsed.

I can definitely work on this, I'm just wondering whether this belongs here or in its own project.

Thanks!

kpdecker:
I'm fine with it in the core library as long it is not included in the
client builds.

kpdecker:
@fivetanley are you bailing on this goal in addition to the PR?

fivetanley:
Changing direction a bit. Will probably be a standalone library similar to benjamn/recast.

zeppelin:
@fivetanley I wrote something like this for Handlebars: https://github.com/zeppelin/handlebars-serializer

Dunno if that helps, or you need any help regarding the same feature for HTMLBars

zeppelin:
It is by no means feature-complete, but it does cover my use cases: https://github.com/zeppelin/handlebars-serializer/blob/master/packages/handlebars-serializer/tests/serializer-test.js

satyajitnayk:
You can Try https://github.com/handlebars-lang/handlebars.js/issues/169#issuecomment-1772483796

PR #936, opened 2015-01-07
store quote type on the StringLiteral AST instance

fivetanley:
Changes the Grammer to pass DOUBLE_QUOTED_STRING or
SINGLE_QUOTED_STRING instead of STRING, so that information
about the quote type can be passed to the AST nodes.

This can be useful in reprinting an AST without changing
the style of the original template author.

This is part of https://github.com/wycats/handlebars.js/issues/935

### DIFF
--- a/lib/handlebars/compiler/ast.js
+++ b/lib/handlebars/compiler/ast.js
@@ -73,11 +73,14 @@ var AST = {
     this.depth    = depth;
   },
 
-  StringLiteral: function(string, locInfo) {
+  StringLiteral: function(string, locInfo, quoteLiteral) {
     this.loc = locInfo;
     this.type = 'StringLiteral';
     this.original =
       this.value = string;
+    this.isSingleQuoted = quoteLiteral === "'";
+    this.isDoubleQuoted = !this.isSingleQuoted;
+    this.quoteType = this.isSingleQuoted ? 'single' : 'double';
   },
 
   NumberLiteral: function(number, locInfo) {
--- a/spec/ast.js
+++ b/spec/ast.js
@@ -93,6 +93,18 @@ describe('ast', function() {
       var string = new handlebarsEnv.AST.StringLiteral("6", LOCATION_INFO);
       testLocationInfoStorage(string);
     });
+
+    it('stores whether it was a double quoted or single quoted string', function(){
+      var string = new handlebarsEnv.AST.StringLiteral("6", LOCATION_INFO, '"');
+      equals(string.isDoubleQuoted, true);
+      equals(string.isSingleQuoted, false);
+      equals(string.quoteType, 'double');
+
+      string = new handlebarsEnv.AST.StringLiteral("6", LOCATION_INFO, '\'');
+      equals(string.isDoubleQuoted, false);
+      equals(string.isSingleQuoted, true);
+      equals(string.quoteType, 'single');
+    });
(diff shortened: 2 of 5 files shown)


fivetanley:
Weird that NVM is failing...

fivetanley:
Closing as @mmun has convinced me this is a rabbit hole if we can't preserve whitespace.

fivetanley:
@kpdecker I'm reopening as I got some guidance from Yehuda and Rob from the Ember Core Team from their recent work on HTMLBars. I think the printer is more feasible than I thought since the location info is already stored for all(?) node types.

fivetanley:
Do you think we can introduce this in the 3.x series? The AST remains semver compatible but the tokens do not. I don't know how much people use tokens vs the AST.

mmun:
I don't understand the point of this. If you're going to write a re-printer you're going to need to use location info + the original string input in order to preserve whitespace. In particular, you can already see the quotes by inspecting the location info of a string literal + the original string. What does this buy you?

I guess I am OK with storing the raw value on the AST since esprima does it for JS (probably for a good reason). e.g. `'foo'; 0; null; undefined; 5.6e7;` parses as

```json
{
    "type": "Program",
    "body": [
        {
            "type": "ExpressionStatement",
            "expression": {
                "type": "Literal",
                "value": "foo",
                "raw": "'foo'"
            }
        },
        {
            "type": "ExpressionStatement",
            "expression": {
                "type": "Literal",
                "value": 0,
                "raw": "0"
            }
        },
        {
            "type": "ExpressionStatement",
            "expression": {
                "type": "Literal",
                "value": null,
                "raw": "null"
            }
        },
        {
            "type": "ExpressionStatement",
            "expression": {
                "type": "Identifier",
                "name": "undefined"
            }
        },
        {
            "type": "ExpressionStatement",
            "expression": {
                "type": "Literal",
                "value": 56000000,
                "raw": "5.6e7"
            }
        }
    ]
}
```

kpdecker:
@fivetanley I'm not a huge fan of adding parameters after the `locInfo` parameter but that is mostly an OCD concern relative to all of the other constructors that have it as the last parameter.

That being said, I think @mmun brings up an interesting point. Rather than complicating the parser (each  new production can add a fair amount of overhead), what if we added a raw value and move the escape handling into the `StringLiteral` constructor? I.e. something like

```
StringLiteral: function(string, locInfo) {
  if (/^"(\\"|[^"])*"$/.test(string)) {
    this.raw = string;
    string = strip(1,2).replace(/\\"/g,'"');
  } else if (/^'(\\'|[^'])*'$/.test(string)) {
    this.raw = string;
    string = strip(1,2).replace(/\\'/g,"'");
  }
...
```

And then in the tokenizer we would simplify to:

```
<mu>'"'("\\"["]|[^"])*'"'        return 'STRING';
<mu>"'"("\\"[']|[^'])*"'"        return 'STRING';
```

(Potentially either of these could be simplified further but I'm not sure how much O want to further tempt the regex gods)

This would break for cases where people are passing in a string that happens to parse to a string literal syntax so we might want to do this as a 4.x but I have no problem rolling a 4.x as long as we continue to document what the changes are and how people have to deal with them. Given the amount of hell I catch on every release I have no problem doing majors more.

fivetanley:
sounds good

PR #938, opened 2015-01-14
Fix example using #with helper

diwo:
Example using `#with` helper should close tag using `{{/with}}`


kpdecker:
Thanks!

kpdecker:
Released in 3.0.0

PR #940, opened 2015-01-17
Add missing reserved words so compiler knows to use array syntax:

mattflaschen:
The added reserved words are:

* await
* null
* true
* false

IE 8 was failing to compile Handlebars-generated source code
because it had helpers.null.

I came up with this list by diffing
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Keywords
against the ones Handlebars already had.

### DIFF
--- a/lib/handlebars/compiler/javascript-compiler.js
+++ b/lib/handlebars/compiler/javascript-compiler.js
@@ -1025,7 +1025,8 @@ var reservedWords = (
   " class float package throws" +
   " const goto private transient" +
   " debugger implements protected volatile" +
-  " double import public let yield"
+  " double import public let yield await" +
+  " null true false"
 ).split(" ");
 
 var compilerWords = JavaScriptCompiler.RESERVED_WORDS = {};
--- a/spec/javascript-compiler.js
+++ b/spec/javascript-compiler.js
@@ -19,6 +19,12 @@ describe('javascript-compiler api', function() {
       };
       shouldCompileTo("{{foo}}", { bar_foo: "food" }, "food");
     });
+
+    // Tests nameLookup dot vs. bracket behavior.  Bracket is required in certain cases
+    // to avoid errors in older browsers.
+    it('should handle reserved words', function() {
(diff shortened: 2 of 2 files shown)


kpdecker:
Can you add a test case for this. We run our tests in Sauce including on old IE so that would prevent regression in the future.

mattflaschen:
Sure.  It's now testing both the dot syntax and bracket syntax paths.

kpdecker:
That assert require is going to break when tested in the actual browsers themselves. I was thinking something more end to end.

```
    shouldCompileTo('{{null}}', { 'null': 'foo' }, 'foo');
```

(Or whatever the end user behavior that you say may be.

mattflaschen:
Sorry, I forgot the test would be run in the browsers.

I've changed it to an end-to-end test.  We were using {{~null~}} as a way of trimming (I realize there are other ways to do that), so that's what I've done here too.

kpdecker:
Thanks!

Could you elaborate more on the trimming? I'm not sure why you'd want to create these additional selectors as it seems like `{{foo~}}` does the same thing without the overhead. (Regardless of that question, I've merged the PR)

mattflaschen:
An example is https://git.wikimedia.org/blob/mediawiki%2Fextensions%2FFlow.git/2b74fa3f4088a0b61144b10f08d9b93fc0945ea5/handlebars%2Fflow_moderation_actions_list.handlebars#L4 .

I didn't write that part (though I've worked on trimming stuff with Handlebars elsewhere in our codebase).  However, I think it's used because there is no Handlebars {{ entity at that location (on either side), but the trimming functionality is nonetheless wanted (so there is no space in the li before the a).

FYI, I replaced it with noop, as I think that is a little more obvious, and so we don't need to upgrade immediately (but thank you for reviewing promptly).

kpdecker:
Gotcha. f30b3ea3293ad28bad8167e40b8372c48cb2419e allows for the same with zero overhead after compilation, but that has not yet been released (and apparently doesn't support `~}}`, which I need to fix) so noop is your best option right now.

kpdecker:
Released in 3.0.0

PR #941, opened 2015-01-18
Add support for dynamic partial names

### LINKED ISSUE #933: Partials don't work with subexpressions (like they should?)
It seems to me, that according to [src/handlebars.yy](https://github.com/wycats/handlebars.js/blob/master/src/handlebars.yy#L80), the partial tag (`{{> ...}}`) should support subexpressions: `OPEN_PARTIAL sexpr CLOSE ...`

But with a freshly checked out handlebars repository, and `testme.js` in the same folder:

```javascript
// testme.js
var Handlebars = require('./');

var source = "{{> (subexpr)}}";
Handlebars.registerPartial('partial', 'this is a partial')
var template = Handlebars.compile(source);

var data = { "subexpr": "partial" };
var result = template(data);
```

`node testme.js` gives:

```
/tmp/handlebars.js/dist/cjs/handlebars/compiler/parser.js:124
    throw new Error(str);
          ^
Error: Parse error on line 1:
{{> (subexpr)}}
----^
Expecting 'STRING', 'NUMBER', 'ID', 'DATA', got 'OPEN_SEXPR'
    at Object.parseError (/tmp/handlebars.js/dist/cjs/handlebars/compiler/parser.js:124:11)
    at Object.parse (/tmp/handlebars.js/dist/cjs/handlebars/compiler/parser.js:176:22)
    at HandlebarsEnvironment.parse (/tmp/handlebars.js/dist/cjs/handlebars/compiler/base.js:25:30)
    at compileInput (/tmp/handlebars.js/dist/cjs/handlebars/compiler/compiler.js:448:19)
    at ret (/tmp/handlebars.js/dist/cjs/handlebars/compiler/compiler.js:457:18)
    at Object.<anonymous> (/tmp/handlebars.js/testme.js:8:14)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
```

Is this a bug or a comprehension-error on my part?

kpdecker:
Uses the subexpression syntax to allow for dynamic partial lookups. Ex:

```
{{> (helper) }}
```

Fixes #933

### DIFF
--- a/docs/compiler-api.md
+++ b/docs/compiler-api.md
@@ -204,7 +204,7 @@ function ImportScanner() {
 ImportScanner.prototype = new Visitor();
 
 ImportScanner.prototype.PartialStatement = function(partial) {
-  this.partials.push({request: partial.sexpr.original});
+  this.partials.push({request: partial.name.original});
 
   Visitor.prototype.PartialStatement.call(this, partial);
 };
--- a/docs/compiler-api.md
+++ b/docs/compiler-api.md
@@ -115,6 +115,17 @@ interface SubExpression <: Expression {
 
 `isHelper` is not required and is used to disambiguate between cases such as `{{foo}}` and `(foo)`, which have slightly different call behaviors.
 
+```java
+interface PartialExpression <: Expression {
+    type: "PartialExpression";
+    name: PathExpression | SubExpression;
+    params: [ Expression ];
+    hash: Hash;
+}
+```
+
+`path` may be a `SubExpression` when tied to a dynamic partial, i.e. `{{> (foo) }}`
+
 ##### Paths
 
 ```java
@@ -220,6 +231,8 @@ The `Handlebars.JavaScriptCompiler` object has a number of methods that may be c
   - `parent` is the existing code in the path resolution
   - `name` is the current path component
   - `type` is the type of name being evaluated. May be one of `context`, `data`, `helper`, or `partial`.
+
+  Note that this does not impact dynamic partials, which implementors need to be aware of. Overriding `VM.resolvePartial` may be required to support dynamic cases.
 
 - `depthedLookup(name)`
   Used to generate code that resolves parameters within any context in the stack. Is only used in `compat` mode. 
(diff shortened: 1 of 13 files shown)


kpdecker:
@mmun Would love your feedback here as this introduces AST changes, but now is the time to make them.

mmun:
I don't think introducing a new Expression is the solution here. A PartialExpression should be substitutable for an Expression but they don't make sense inside a Param or Hash value in a mustache, (you can't get there syntactically, but leaving it as a possible AST seems bad).

Maybe just extend the the `partial` production to understand nested sexprs in the first slot?

```
partial
  : OPEN_PARTIAL OPEN_SEXPR sexpr CLOSE_SEXPR param? hash?  CLOSE -> ...
  | OPEN_PARTIAL sexpr CLOSE -> new yy.PartialStatement($2, yy.stripFlags($1, $3), yy.locInfo(@$))
  ;
```

I am starting to think that sexpr should include `OPEN_SEXPR` and `CLOSE_SEXPR` and each of the the statements should explicitly list out the inputs they accept.

kpdecker:
@mmun updated the AST spec doc to remove direct SubExpression references in statements. This simplifies a few things, but complicates others. LMK if we are we on the same page there and I'll update the code to follow this.

mmun:
:+1: :+1: This seems very reasonable to me.

kpdecker:
They become more complicated, but they are duck typing data that was already there. The way that I'm thinking about coding it, it's actually less complicated in the compiler (no intermediate sexpr block for the same content) at the cost of potentially having more complicated parser.

Will ping this review after the code changes are done.

kpdecker:
@mmun code changes pushed, if you are concerned about that. Looks like the library size only goes up by ~600bytes, which isn't too bad. Note that I did go the route of adding support for `{{@foo bar}}` as we chatted about briefly off thread.

mmun:
I read it over. Looks good.

zordius:
:+1:

kpdecker:
Released in 3.0.0

jonschlinkert:
this is fantastic, :+1:

PR #957, opened 2015-02-07
Boolean literals can be passed to mustaches

marcioj:
In [htmlbars](https://github.com/tildeio/htmlbars), I believe the following component `<x-foo bool=true ></ x-foo>` is transformed to `<x-foo bool={{true}} ></ x-foo>`. But mustaches with boolean literals don't work in handlebars, just numbers. This PR fixes this.

//cc @mmun

### DIFF
--- a/lib/handlebars/compiler/compiler.js
+++ b/lib/handlebars/compiler/compiler.js
@@ -172,11 +172,8 @@ Compiler.prototype = {
   },
 
   MustacheStatement: function(mustache) {
-    if (!mustache.path.type.match(/Literal$/)) {
-      this.SubExpression(mustache);
-    } else {
-      this.accept(mustache.path);
-    }
+    transformLiteralToPath(mustache);
+    this.SubExpression(mustache);
 
     if(mustache.escaped && !this.options.noEscape) {
       this.opcode('appendEscaped');
@@ -487,5 +484,12 @@ function argEquals(a, b) {
       }
     }
     return true;
+  }
+}
+
+function transformLiteralToPath(mustache) {
+  if (mustache.path.type.match(/Literal$/)) {
+    var literal = mustache.path;
+    mustache.path = { type: 'PathExpression', original: String(literal.original), parts: [String(literal.original)], depth: 0, data: false };
   }
 }
--- a/lib/handlebars/compiler/compiler.js
+++ b/lib/handlebars/compiler/compiler.js
@@ -172,7 +172,11 @@ Compiler.prototype = {
   },
 
   MustacheStatement: function(mustache) {
-    this.SubExpression(mustache);
+    if (!mustache.path.type.match(/Literal$/)) {
+      this.SubExpression(mustache);
+    } else {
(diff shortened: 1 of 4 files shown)


kpdecker:
I'd like to see end to end tests that demonstrate what the user can expect in Handlebars if they use this construct in addition to just parser tests.

kpdecker:
Thanks. I remember having a conversation regarding this with @mmun a few months ago. At the time I think my conclusion was that `{{true}}` for example should evaluate the field named true, but it wasn't that solid of a conclusion. Right now I'm not sure what the best behavior is here as:
1) Evaluating literal values as helpers feels a bit odd
2) Outputting literals as mustache values doesn't make a lot of sense either, from the perspective of Handlebars proper.

Can someone outline the need for this in htmlbars for the sake of public record?

mmun:
This feature would let HTMLBars deal with passing literals for data binding, e.g. `<my-component enabled={{true}}>`. HTMLBars will share the parser of Handlebars, but have modified semantics in this case (though we try to align as much as possible).

@marcioj You should add additional tests that ensure that `{{true}}` has similar semantics to `{{3.14}}`. I think you won't actually need to change the `MustacheStatement` visitor.

mmun:
Not too sure about `{{"foo"}}`. It's an edge case I'm not too worried about. I think it's supposed to do a property look up as well though.

marcioj:
Thank you for your replies!

@kpdecker I agree that from the perspective of handlebars this is not so useful. I believe that rarely people will use `{{true}}` in your handlebars templates. Like @mmun described, this changes are more related to make the template be parsed in htmlbars, since it uses the handlebars parser. Of course handlebars needs some behavior, so what comes to my mind was just output the primitive value in the html.

> You should add additional tests that ensure that {{true}} has similar semantics to {{3.14}}

@mmun can you point out what semantic isn't covered? Isn't clear to me.

>  I think you won't actually need to change the MustacheStatement visitor.

I changed the MustacheStatement because `mustache.path` can be a literal and `this.SubExpression(mustache);` throws in these cases. I think `mustache.path` being a literal seems wrong, but I don't know other way to solve this.

> Not too sure about {{"foo"}}. It's an edge case I'm not too worried about. I think it's supposed to do a property look up as well though.

I agree that is a edge case, I just added the test for {{"foo"}} to keep the behavior consistent.

mmun:
I would assume that in Handlebars {{true}}, {{"foo"}} and {{10}} should all do property lookups for the keys "true", "foo" and "10".

marcioj:
I updated based on your feedback @mmun
This is my first PR to handlebars, so please let me know if I am missing something important here.

zordius:
Extra reminds: here are 2 known ways to access properties when it's something like boolean: `{{./true}}` , `{{[true]}}` .

kpdecker:
I updated this to handle a few missing cases and have merged. This should be released shortly.

kpdecker:
Released in 3.0.0

marcioj:
Thank you!

Issue #887, opened 2014-10-21
Handlebars.noConflict() option?


bradvogel:
It would be useful to have a noConflict() method to pull Handlebars out of the global namespace, similar to jQuery's http://api.jquery.com/jquery.noconflict/. It would be used like:

```
(function(){
    var hb = Handlebars.noConflict(); // Removes window.Handlebars from the global namespace.
})();
```

Use cases for this:
1. You're writing a gmail chrome extension that shares global namespace with gmail. You want to use Handlebars but don't want window.Handlebars.
2. You're using Handlebars inside of an environment that already declares it (such as Meteor does for backwards compatibility).

kpdecker:
Released in 3.0.0

bernland:
Can you please give an example?

    var hb = Handlebars.noConflict();
    hb.compile(...);

results in the following error: *TypeError: hb is undefined*.

kpdecker:
It does not return the instance as currently implemented, so you would need to do something like

```
var hb = Handlebars;
Handlebars.noConflict();
```

Glad to accept a pull request for the return value though, it's minimal overhead.

PR #959, opened 2015-02-11
Return value for Handlebars.noConflict();

bernland:
As per our discussion #887, it would be great if the noConflict() method returned an instance of Handlebars.

The docs at http://handlebarsjs.com/reference.html (see paragraph Handlebars.noConflict()) already suggests it does.


kpdecker:
I think you selected the wrong branch when creating your pull request. Can you reopen against your own branch?

bernland:
This is really embarrassing: I forked your project into my account, then 
tried to create a new pull request.

Error message: There isn't anything to compare wycats:master and 
bernland:master are identical.

What am I missing here?

kpdecker:
General flow that I use when submitting a PR from a fork:

```
git fetch upstream
git checkout upstream/master
git checkout -b $featureBranchName
```

You then make all of the changes that you need, push, and open the the PR. Upstream here is the remote for the source repo, for handlebars you'd need to set that up, something like:

```
git remote add upstream git@github.com:wycats/handlebars.js.git
```

PR #984, opened 2015-03-19
Adding documentation for passing arguments into partials

johneke:
Great feature!!! However I only found it on stack overflow, pull requests, and github issues after spending lots of time searching for how to do this.

It would be awesome if this was right in the handlebars docs!

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -239,16 +239,20 @@ template(data);
 Partials can also accept parameters
 
 ```js
-var source = "<div>{{> roster people=listOfPeople rosterName}}</div>";
+var source = "<div>{{> roster rosterProperties people=listOfPeople}}</div>";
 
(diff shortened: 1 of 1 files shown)


zordius:
This example cause error:

```
Uncaught Error: Parse error on line 1:
...tOfPeople rosterName}}
-----------------------^
Expecting 'EQUALS', got 'CLOSE'
```

I tried it with http://builds.handlebarsjs.com.s3.amazonaws.com/handlebars-ab96073.js

When I change source to `{{> roster people=listOfPeople}}` , it works well.

johneke:
Thanks @zordius!! I just wanted to show an example which had both a hash parameter and a context parameter.

I've updated :)

kpdecker:
Released in 3.0.1

Issue #855, opened 2014-09-02
Update to es6 module syntax


MajorBreakfast:
`module myModule from 'my-module'` became `import * as myModule from 'my-module'`
http://jsmodules.io/

I think many of you are already aware of that. I'm just posting it as an issue because it needs to be addressed.

kpdecker:
Released in 3.0.2

PR #993, opened 2015-04-09
Use `import` instead of `module`

seanpdoyle:
* export Helpers as default export in `handlebars/compiler/base`

### DIFF
--- a/lib/handlebars/compiler/base.js
+++ b/lib/handlebars/compiler/base.js
@@ -1,12 +1,29 @@
 import parser from "./parser";
 import AST from "./ast";
 import WhitespaceControl from "./whitespace-control";
-module Helpers from "./helpers";
+import {
+  SourceLocation,
+  stripFlags,
+  stripComment,
+  preparePath,
+  prepareMustache,
+  prepareRawBlock,
+  prepareBlock
+} from "./helpers";
 import { extend } from "../utils";
 
 export { parser };
 
 var yy = {};
+var Helpers = {
+  SourceLocation: SourceLocation,
+  stripFlags: stripFlags,
+  stripComment: stripComment,
+  preparePath: preparePath,
+  prepareMustache: prepareMustache,
+  prepareRawBlock: prepareRawBlock,
+  prepareBlock: prepareBlock
+};
 extend(yy, Helpers, AST);
 
 export function parse(input, options) {


seanpdoyle:
@mmun would you mind merging this into your branch so we can move forward with https://github.com/emberjs/ember.js/pull/10853

stefanpenner:
`import * as helpers from '...'`

seanpdoyle:
@stefanpenner 

```js
import * as Helpers from "./helpers";
```

results in 

```
Warning: Line 4: Unexpected token * Use --force to continue.
```

The toolchain doesn't support that syntax

kpdecker:
Can you explain the reason for this change?

kpdecker:
Looks like this is trying to hack around #855 but is introducing more tech debt. Not really a fan of this but would love to hear the argument for it.

seanpdoyle:
@kpdecker you're right about that.

I'd certainly prefer to use `import * as Helpers from "./helpers";`.

This change is motivated by https://github.com/emberjs/ember.js/pull/10853#issuecomment-91089580

What would it take to support that syntax (given the change in ES6 landscape since #855)

stefanpenner:
> import * as Helpers from "./helpers";

it is valid, we will need to upgrade the transpiler to support this (now or later)

seanpdoyle:
@kpdecker could we update https://github.com/kpdecker/es6-module-packager to use babel?

kpdecker:
That could be an option. es6-module-transpiler was a no go as it generated code that wouldn't run under IE. I don't have time to look into that now but I'd be glad to look at a pull request over there.

stefanpenner:
Use the es6modules transpiler we have in the ember cli repo. It's Esperanto under the hood with flags enabled or enable able for es3 safety. It is also ultimately what ember core is currently build with

seanpdoyle:
@stefanpenner do you mean https://github.com/ember-cli/broccoli-es6modules?

Does that introduce a broccoli dependency to the build process?

stefanpenner:
If you don't want to do that. Look at how it calls Esperanto under the hood and drop Esperanto in here. There may even be an appropriate plugin for this build system.

Just make sure to call Esperanto with the same flags ember-cli does. So es3 safety requirement is met.

kpdecker:
Closing in favor of #998

seanpdoyle:
:clap:

PR #1008, opened 2015-04-25
Added handlebars-layouts to the in-the-wild list

shannonmoeller:
I've written a set of helpers which implement layout blocks similar to Jade, Jinja, Swig, Twig, and others. It seems to be gaining some popularity (around 10k npm downloads per month at present) and I would be honored to have it included in the in-the-wild list in case it may prove useful to others.


shannonmoeller:
Thank you kindly!

kpdecker:
Released in 3.0.3

PR #1015, opened 2015-05-01
Fix tokenizer test for double quoted strings

aymerick:
There are two consecutive tests with the same input data:

    {{ foo bar \'baz\' }}

I suppose the first test should be about testing double quoted string so I changed its input data to:

    {{ foo bar \"baz\" }}


kpdecker:
Batch style replace regression. Thanks for the fix!

kpdecker:
Released in 4.0.0

PR #1054, opened 2015-07-07
Compile failing when partial is an instance of Handlebars.SafeString

rd5:
When partial's `indent` option is used, runtime compilation of string partial is failing for partials produced by `Handlebars.SafeString` because of `split` method call.

```js
let lines = result.split('\n');
```
https://github.com/wycats/handlebars.js/blob/master/lib/handlebars/runtime.js#L52

See [jsfiddle demo](http://jsfiddle.net/j1j60b0u/) for details.

I have fixed it in a straight way by explicitly calling `result.toString()`.

### DIFF
--- a/lib/handlebars/runtime.js
+++ b/lib/handlebars/runtime.js
@@ -49,7 +49,7 @@ export function template(templateSpec, env) {
     }
     if (result != null) {
       if (options.indent) {
-        let lines = result.split('\n');
+        let lines = result.toString().split('\n');
         for (let i = 0, l = lines.length; i < l; i++) {
           if (!lines[i] && i + 1 === l) {
             break;


kpdecker:
Can you add a test to ensure that this does not regress and I'll be glad to
add. spec/regressions.js might be a good place to add this, otherwise
spec/partials.js

rd5:
I'm little bit unsure about exact solution to this problem. Should partial always return a string only?  
If so it'll be better to call result.toString() inside `invokePartial` and not in the any conditions later, e.g. one in `invokePartialWrapper`.  
Could you tell me about partial's main concept? Is it just a part of the template (and in that case it _should_ always return string), or partial in Handlebars is something more than just a piece of template (and maybe it should return objects like SafeString in some cases)?

kpdecker:
You raise an interesting point. I was surprised that you were able to cause a safe string return from a partial in the first place. Modifying your example a little bit, http://jsfiddle.net/j1j60b0u/1/, shows that a template with a simple wrapper can return a SafeString as well, which could lead to issues down the road. I think we need to make sure that the return from any template, partial or not, is always a string and have proper tests to ensure that :)

kpdecker:
@rd5, I went the route of consistent string return from the API and implemented your test case there. Thanks for the report!

rd5:
Thank you!

PR #1055, opened 2015-07-07
Update HelperExpression to be able to evaluate nodes without params a…

chrisrink:
Update HelperExpression to be able to evaluate nodes without params attribute. 

I want to be able to traverse the entire AST and check if each node/param/hash is a helper or not. This currently blows up on literals that don't have params.

### DIFF
--- a/lib/handlebars/compiler/ast.js
+++ b/lib/handlebars/compiler/ast.js
@@ -131,7 +131,7 @@ let AST = {
     // * it is an eligible helper, and
     // * it has at least one parameter or hash segment
     helperExpression: function(node) {
-      return !!(node.type === 'SubExpression' || node.params.length || node.hash);
+      return !!(node.type === 'SubExpression' || (node.params && node.params.length) || node.hash);
     },
 
     scopedId: function(path) {


chrisrink:
Actually, this is not right either as a Helper could be a MustacheStatement or BlockStatement. I did notice that the path object of the statement seems to only exist for helper statements. So I modified it as such: 

```javascript
 return !!((statement.path && (statement.type === 'SubExpression' || statement.params && statement.params.length) || statement.hash));
```

Seems a little hacky though so there might be a more concise way to handle this.

kpdecker:
Please update the PR with what you'd like to see and include test cases. I'm all for bulletproofing the code where it's making unsafe assumptions.

kpdecker:
(Comment here when you've updated as otherwise we might not see the changes)

kpdecker:
@chrisrink thanks for the initial work here. Just landed a patch the fixes this in master.

chrisrink:
Great news! I think this should work. One last thing to check. I have actually been using the below statement. 

{code}
        return !!((statement.path && statement.path.original) && (statement.type === 'SubExpression' || (statement.params && statement.params.length) || statement.hash));
{code}
And this seems to be working for me. So I just wanted to double check that the statement.path and statement.path.original parts would be caught in Mustache and Block statement calls.

Issue #1042, opened 2015-06-18
Use of `as ||` in `with`


agladysh:
I'd like to create an alias to outer `this` in my template.

```
{{#with this as |foo|}
  {{#each bar}}
    Foo: {{foo.baz}} 
    bar: {{bar}}
  {{/each}}
{{/with}
```

But the code above doesn't work... What am I doing wrong? Is `each` the only standard template that supports the `as ||` feature?

Fiddle: http://jsfiddle.net/xmeq3x76/1/

kpdecker:
`with` does not support block parameters right now but this seems like a useful feature to have

Issue #1021, opened 2015-05-08
The each-helper ignores the Empty-String property


stephangiesen:
Hello,
thank you for this nice tool, it helped us a lot.

I'm stumbled on a problem with the each-helper.

It ignores the Empty-String property. When you iterate thru an object like this
```javascript
{
 "": "Please Select" , 
 "adult": "Adult", 
 "teenager": "Adolescents aged 13 - 18 years", 
 "child": "Children under 12 years"
}
```
The first option with the empty string as key will be ignored.
To fix this I changed in handlebars.runtime-v3.0.3.js the two lines 325 and 332 from

```javascript
if (priorKey) {
```
to
```javascript
if (priorKey !== undefined) {
```

So it works fine for me.

Nice regards 
Stephan

kpdecker:
Thanks for the report. Will have this fixed by the next release.

PR #1059, opened 2015-07-14
Implemented block params support for `each` and `with`

ericbn:
Block parameter does not change current context

### DIFF
--- a/spec/builtins.js
+++ b/spec/builtins.js
@@ -136,14 +136,28 @@ describe('builtin helpers', function() {
       equal(result, '0. goodbye! 0 1 2 After 0 1. Goodbye! 0 1 2 After 1 2. GOODBYE! 0 1 2 After 2 cruel world!', 'The @index variable is used');
     });
 
+    it('each should change context', function() {
+      var string = '{{#each foo}}{{.}}{{bar}}{{@key}} {{@index}} {{@first}} {{@last}} {{baz}}{{/each}}';
+      var hash = {foo: ['a', 'b', 'c'], bar: 'd', baz: 'e'};
+      shouldCompileTo(string, hash, 'a0 0 true false b1 1 false false c2 2 false true ');
+    });
+
     it('each with block params', function() {
-      var string = '{{#each goodbyes as |value index|}}{{index}}. {{value.text}}! {{#each ../goodbyes as |childValue childIndex|}} {{index}} {{childIndex}}{{/each}} After {{index}} {{/each}}{{index}}cruel {{world}}!';
+      var string = '{{#each goodbyes as |value index|}}{{index}}. {{value.text}}! {{#each goodbyes as |childValue childIndex|}} {{index}} {{childIndex}}{{/each}} After {{index}} {{/each}}{{index}}cruel {{world}}!';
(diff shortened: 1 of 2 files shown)


kpdecker:
These look like large changes to existing features. Can you provide a summary of the changed and new behaviors?

ericbn:
Sure, @kpdecker.

Summary of changes:
* Implemented block params support for `with` block (this was requested in #1042)
* Changed current block params support implementation for `each` to make it "context-preserving". The same context-preserving behavior applies to the `with` with block params implemented above.
* Fixed the `each` helper so it also iterates over an empty string property in objects (as requested in #1021).
* Added unit tests for all corrections and new implementations.

Now, the context-preserving behavior for blocks with block parameters comes from the Ember specifications of the block params feature. Refences about this can be found in the corresponding [RFC](https://github.com/emberjs/rfcs/pull/3#issuecomment-61543805):

>  Introduce block parameters to the Handlebars language to standardize context-preserving helpers ([ref](https://github.com/emberjs/rfcs/pull/3#issue-40512558))

This idea of a context-preserving block was already introduced to Ember in their custom `{{#each ... in ...}}` and `{{#with ... as ...}}`: ", and the generic block param specification aims to make it standard:

 > `{{#each person in persons}}` was designed to not lose the parent context. It works, but its a hack which should be standardized ([ref](https://github.com/emberjs/rfcs/pull/3#issuecomment-57810150))

The [Ember 1.10 release notes](http://emberjs.com/blog/2015/02/07/ember-1-10-0-released.html#toc_block-params) also gives a summary about it.

spec/builtins.js
```diff
@@ -47,6 +47,22 @@ describe('builtin helpers', function() {
       var string = '{{#with person}}Person is present{{else}}Person is not present{{/with}}';
       shouldCompileTo(string, {}, 'Person is not present');
     });
```
ericbn: @kpdecker, I hope the specs can also serve as a more detailed summary of the changes.

ericbn:
There is a change I did in the `each` helper, when block params are used, that deserves also a explanation.

The block params for `each` are not only the iteration item (first param), but also key, index, first and last (in this order).

    {{#each items as |item key index first last|}}
    {{/each}}

Also,  when block params are used, no `data` attribute as `@key` or `@index` is generated, so not to conflict with the current context, as no context switching occurs.

    {{#each students}}
        {{#each exams as |exam, examKey|}}
            
            Student name: {{name}} 
            Student key: {{@key}}
            Exam key: {{examKey}}
        {{/each}}
    {{/each}}

I believe creating a new `data` frame while not switching context would break the engine anyway...

kpdecker:
Ember and handlebars have distinct behaviors on many levels. I can't accept
a breaking change like the context one you are proposing. The users will
lynch me if I make breaking changes that are so integral for no reason
other than parity with a project that we never had parity with. I'm glad to
accept smaller changes in distinct PRs but this is too big and too breaking
for me to merge as is.

(Sorry for the quick response, I'm traveling and wanted to provide some
feedback so this doesn't sit without comment for weeks)

ericbn:
Thank you for the quick feedback, @kpdecker.

Issue #1018, opened 2015-05-05
Inline Partials


kpdecker:
# Inline Partials

Inline partials provide a syntax that allows partials and helpers to have named programs that they can execute at will. This allows for use cases such as local partials, layout templates, and helpers that have more conditional rendering options than just positive and negative cases.

## Defining

Inline partials are defined using the `<` flag.

```
{{< bar}}
{{/bar}}
```

Will declare the `bar` inline partial which will be made available when the parent runs.

Each block has it's own set of named inlined partials and are able to access any inline partials defined in parent scopes.

Inline partials follow the same standalone rules as blocks.

Multiple inline partials with the same name in a given block result in a compile error.

Inline partials have access to all block parameters defined by their parents and may also define their own:

```
{{< bar as |foo| }}
  {{foo}}
{{/bar}}
```

## Calling from helpers

Inline partials contained within helper calls will be exposed to the call's `options.partials` object as a named value.

```
{{#foo}}
  {{*bar}}
    Blah
  {{/bar}}
{{/foo}}
```

```
function foo(options) {
  return options.partials.bar(context);
}
```

The existence of a given inline partial is not guaranteed so helpers should implement checks when resiliency is a concern.

## Calling from templates

Within templates, inline partials may be called using the normal partial syntax:

```
{{< foo}}bar{{/bar}}
{{< baz}}{{> foo}}{{/baz}}
{{> baz}}
```

Custom contexts or parameters may be passed to a inline partial in the same manner as other partials:

```
{{> bar foo baz=bat}}
```

Would execute the inline partial `bar` with context `foo`, augmented with the value `baz` set to `bat`.

Inline partials have any block parameters from their defining scope made available to them at execution time.

When run using `{{> }}`, the inlined partials take priority over higher scoped partials that may be defined.

### Partial Blocks

Partial calls may now have a child block, which define inline partials that may be passed to the executed partial as well as a default behavior to be executed if a given partial does not exist.

```
{{#> foo}}
  {{*bar}}
    Blah
  {{/bar}}
{{/foo}}
```

Would call the partial `foo`, defining the `bar` inline partial which will be passed to the partial, which can execute it as follows:

```
{{> bar}}
```

Any content in the partial block other than inline partials is not executed if a given partial exists. If the partial would like to render this content, to augment default behavior for example, it may do so using the `@default` partial name.

```
{{> bar}}
  foo
{{/bar}}
```

```
{{< bar}}
  {{> @default}}
  other foo
[...]

PR #1076, opened 2015-08-14
Implement partial blocks

kpdecker:
This allows for failover for missing partials as well as limited templating ability through the `{{> @partial-block }}` partial special case.

Partial fix for #1018

### DIFF
--- a/docs/compiler-api.md
+++ b/docs/compiler-api.md
@@ -83,7 +83,19 @@ interface PartialStatement <: Statement {
     name: PathExpression | SubExpression;
     params: [ Expression ];
     hash: Hash;
-    
+
+    indent: string;
+    strip: StripFlags | null;
+}
+
+interface PartialBlockStatement <: Statement {
+    type: "PartialBlockStatement";
+    name: PathExpression | SubExpression;
+    params: [ Expression ];
+    hash: Hash;
+
+    program: Program | null;
+
     indent: string;
     strip: StripFlags | null;
 }
--- a/docs/compiler-api.md
+++ b/docs/compiler-api.md
@@ -97,7 +97,8 @@ interface PartialBlockStatement <: Statement {
     program: Program | null;
 
     indent: string;
-    strip: StripFlags | null;
+    openStrip: StripFlags | null;
+    closeStrip: StripFlags | null;
 }
 ```
 
--- a/lib/handlebars/compiler/compiler.js
+++ b/lib/handlebars/compiler/compiler.js
@@ -1,3 +1,5 @@
+/* eslint-disable new-cap */
+
 import Exception from '../exception';
 import {isArray, indexOf} from '../utils';
 import AST from './ast';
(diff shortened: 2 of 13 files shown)


kpdecker:
@mmun @wycats could I get your input on this? Trying to carve off pieces from #1018 that can go out in a more standalone fashion vs. a big bang meta feature.

My biggest questions here are:
1. Does the feature generally make sense in isolation?
2. Is the naming of `@partial-block` sensible?
3. Do we need a separate AST node for this or should we merge the two and use the existence of `program` as the differentiator in behavior? This is effectively the way that it's implemented now but it has all of the extra overhead of handling the distinct AST type.

Documentation PR here: https://github.com/wycats/handlebars-site/pull/120

mmun:
Ember doesn't support `{{> }}` constructs so I don't have a lot of experience.

1. Roughly, yes. 
2. It's ok, but I think a better solution is to introduce a {{yield}} + an isolated "partial" AKA components.
3. In this case, having a single AST node seems better if you can get away with it. (With mustache/block you can't because mustaches have an `escaped` field).

kpdecker:
2. I'm not sure what you are suggesting. Could you provide a code example?
3. I was hoping that this could be done, but strip behavior was sufficiently different and there is also something to be said for introducing new types so non-supporting code doesn't have surprises, i.e. the program could be completely ignored, silently, vs. a thrown error.

mmun:
Since the whitespace constraints are different I don't have anything useful to add. Is `@partial-block` implemented via "data" or is it effectively a keyword?

FWIW, I am still interested in the original named blocks proposal.

mmun:
In ember you can write something like

```hbs
{{#each-with-ul people as |p|}}
  Hello {{p.name}}
{{/each-with-ul}}
```

where the `each-with-ul` component is defined in a separate file as

```hbs
<ul>
  {{#each items as |item|}}
    <li>{{yield item}}</li>
  {{/each}}
</ul>
```

This generates the output

```
<ul>
  <li>Hello Kevin</li>
  <li>Hello Yehuda</li>
</ul>
```

I assume Thorax has something similar. The win here is that the scope of the second template above is completely encapsulated so its easier to reason about and easier to reuse. `@partial-block` is similar to `yield`, but with the partial context "leaking" behaviour.

Aligning Handlebars with {{yield}} would also give named blocks a better chance of landing, as they don't make much sense if you can't yield to them.

kpdecker:
I'm having trouble working through your comments about encapsulation and leaking contexts or how this is different from the helper block-based ember example above. Can you help me understand that?

Regarding named blocks, the [decorators](https://github.com/wycats/handlebars.js/compare/partial-block...decorators?expand=1) branch implements a generic way to apply metadata to blocks and wrap in behaviors as well as implements the inline partials behavior that is the end goal. This PR needs to land before can do too much discussion on that, unfortunately.

mmun:
Forget it, the docs give an example that removes my concerns.

```hbs
{{#each children as |child|}}
  {{#> childEntry}}
    {{child.value}}
  {{/childEntry}}
{{/each}}
```

This works as expected, even if your your childEntry partial was something like

```hbs
{{#with whatever as |child|}}
  {{> %block}}
{{/with}}
```

kpdecker:
Sounds like we're mostly in agreement then. Leaving the token name as `@partial-block` and merging as is to unblock the decorators PR and the release in general.

kpdecker:
Released in 4.0.0

Issue #404, opened 2013-01-04
Add named child helpers that can be referenced by block helpers


jhudson8:
It would be nice if we could define named blocks that could be referenced within a Handlebars helper.  I'll use, for example, a special prefix ```{{@``` to defined the child helper.  An example would be:

```
Here is some template content
{{#section}}
  {{*header}}the section header{{/header}}
  {{*body}}the section body{{/body}}
  {{*footer}}the section footer{{/footer}}
{{//section}}
```
And, for example, the nested helpers could be referenced in this way:
```
Handlebars.registerHelper('section', function(options) {
  // it would also be nice to get all child blocks as an array to check for length
  // in case we wanted to do something different (like not require the usage of {{@body}}
  // if no header/footer were present (using special '__children' just as an example)

  var rtn = '<section>';
  if (!options.fn.__children.length) {
    rtn += options.fn();
  } else {
    if (options.fn.header) {
      rtn += ('<div class="header">' + options.fn.header(this) + '</div>');
    }
    rtn += ('<div class="body">' + options.fn.body(this) + '</div>');
    if (options.fn.footer) {
      rtn += ('<div class="footer">' + options.fn.footer(this) + '</div>');
    }
  }
  return new Handlebars.SafeString(rtn);
});
```

kpdecker:
Closing in favor of #1018

PR #1082, opened 2015-08-22
Decorators

kpdecker:
This is intended for merging into master but also dependent on #1076 so PRing against that branch for discussion right now.

This implements inline partials in the form of:

```
{{#*inline "foo"}}bar{{/inline}}
{{> foo }}
```

Which exposes the `foo` template to the current block and all children. This also adds a special case for partial blocks to support template inheritance:

```
{{#> layout}}
  {{#*inline "content"}}bar{{/inline}}
{{/layout}}
```

Which exposes the `content` partial to the `layout` partial, allowing for overrides and other conditional behaviors.

This is implemented through the more generic decorators feature, which are loosely inspired by @wycats's javascript decorators proposal. These allow a declarative means to both annotate particular blocks with metadata as well as to wrap in behaviors when desired, prior to execution. In addition to the inline partial helper case, this allows for blocks to be flagged with metadata, using [this example](https://github.com/wycats/handlebars.js/issues/404#issuecomment-11914641):

```
{{#grid data}}
  {{#*column "Column 1"}}{{foo}}{{/column}}
  {{#*column "Column 2"}}{{bar}}{{/column}}
{{/grid}}
```
```
Handlebars.registerDecorator('column', function(fn, props, container, options) {
  props.columns = props.columns || [];
  props.columns.push({key: options.args[0], body: options.fn});
});
Handlebars.registerHelper('grid', function(data, options) {
  var headers = options.fn.columns.map(function(column) { return column.key; }).join(',');
  return headers + '\n' + data.map(function(data) {
    return options.fn.columns.map(function(column) { return column.body(data); }).join(',');
  }).join('\n');
});
```

/cc @wycats, @mmun, @erisds

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -45,7 +45,7 @@
     "no-extra-boolean-cast": 2,
     "no-extra-parens": 0,
     "no-extra-semi": 2,
-    "no-func-assign": 2,
+    "no-func-assign": 0,
 
     // Stylistic... might consider disallowing in the future
     "no-inner-declarations": 0,
--- a/docs/compiler-api.md
+++ b/docs/compiler-api.md
@@ -276,7 +276,7 @@ The `Handlebars.JavaScriptCompiler` object has a number of methods that may be c
 
   - `parent` is the existing code in the path resolution
   - `name` is the current path component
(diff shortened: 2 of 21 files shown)


kpdecker:
Released in 4.0.0

PR #1083, opened 2015-08-24
Adding a few more badcharacters

FreedomCoder:
It would be interesting to contemplate adding even more characters.

### DIFF
--- a/lib/handlebars/utils.js
+++ b/lib/handlebars/utils.js
@@ -4,11 +4,17 @@ const escape = {
   '>': '&gt;',
   '"': '&quot;',
   "'": '&#x27;',
-  '`': '&#x60;'
+  '`': '&#x60;',
+  '\n' : '\\n',   // NewLine
+  '\r' : '\\n',   // Return 
+  '\b' : '\\b',   // Backspace
+  '\f' : '\\f',   // Form fee
+  '\t' : '\\t',   // Tab
+  '\v' : '\\v',   // Vertical Tab
 };
 
-const badChars = /[&<>"'`]/g,
-      possible = /[&<>"'`]/;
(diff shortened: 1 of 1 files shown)


kpdecker:
This method is intended for escaping HTML content only, not javascript. Could you show me an example of what you are using this for that you'd like this change?

FreedomCoder:
Well, I was adding a few more cases to add common special characters that most libraries include. With that said, I found this when I was testing this edge case  <a href={{foo}}/>   when { 'foo' : 'test.com onload=alert(1)'}  which will be translated as <a href=test.com onload=alert(1)/> which will end on a XSS vulnerability.

kpdecker:
Ping me at kpdecker at gmail with the repo case.

kpdecker:
Sorry that was formatted oddly in my email. No need to ping. This isn't the right fix as you are escaping to a different language and thus creating invalid data. Escaping = would be sufficient to mitigate, no?

FreedomCoder:
I was going to send another patch escaping ‘(‘  , ')’ and ‘=‘.  The reason why I added the other special characters as well is because technically those should be escaped as well, but it could potentially break things if someone is trying to paste something on ```<p>{{text}}</p>``` which has \n. Although that will have to be a design decision from your library. :+1: 

With that said, technically this is more of a fault from developers not inserting the correct “ or ‘ to contain the value of the html Attribute, but it is such a common usage that I think you should try to fix it, by escaping a few more characters as you mentioned and I have recommended above.

kpdecker:
Thanks for bringing this up. I went with the `=` escape as that seemed like the least amount of risk for users. In the compatibility notes I will also note that users should generally opt for quoted attributes whenever possible.

PR #1087, opened 2015-08-30
Fix #each when last object entry has empty key

denniskuczynski:
This PR fixes a minor issue when iterating over an object where the last entry has empty string as its key.

Before the fix, the provided test case would skip the last entry.


kpdecker:
Thanks!

kpdecker:
Released in 4.0.0

PR #1118, opened 2015-10-16
added a basedIndex helper for a human readable index (see https://git…

aaronharding:
Coming from this thread, this commit is for a human readable based index counter. I've been using this in my projects and it seems like others are keen for it too.

https://github.com/wycats/handlebars.js/issues/250


kpdecker:
https://github.com/wycats/handlebars.js/issues/250#issuecomment-148739012

PR #1121, opened 2015-10-19
Include partial name in 'undefined partial' exception message

shinypb:
My project has occasionally had production issues with undefined partials due to programmer error, but we use enough partials that figuring out what's wrong can take a bit of sleuthing.

This patch changes the exception message to include the name of the partial in the error message. This seems in keeping with the general style of the codebase (for example, see various instances in [runtime.js](/wycats/handlebars.js/blob/94c840b/lib/handlebars/runtime.js) — e.g.:
`````js
throw new Exception('The partial ' + options.name + ' could not be found');`)
````

### DIFF
--- a/lib/handlebars/base.js
+++ b/lib/handlebars/base.js
@@ -51,7 +51,7 @@ HandlebarsEnvironment.prototype = {
       extend(this.partials, name);
     } else {
       if (typeof partial === 'undefined') {
-        throw new Exception('Attempting to register a partial as undefined');
+        throw new Exception('Attempting to register a partial called `' + name + '` as undefined');
       }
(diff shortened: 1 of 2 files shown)


kpdecker:
Glad to pull this in but lets go with template strings here (and use `'` or `"` for the embedded identifier). (I believe those are turned on the project options, let me know if they aren't)

shinypb:
Thanks, @kpdecker — I made those changes and appreciate the review. (I tend to shy away from ES6 stuff until asked for it at this point.)

kpdecker:
Looks great. Will merge after travis finishes.

kpdecker:
Released in 4.0.4

Issue #1050, opened 2015-06-30
Splat operator


mixonic:
I've been fielding several requests for a splat operator in Ember templates. Specifically, when using Ember components. For example given:

```handlebars
{{! presuming myHash === { keyA: 'val', keyB: context.someProp } }}
{{my-component *myHash}}
```

It would de-sugar to:

```handlebars
{{my-component keyA="val" keyB=someProp}}
```

This proposal is limited in scope.

* It does not propose a way to de-sugar params from an array, only properties from an object into the hash. I find it useful to consider the params case in designing this API though, presuming that it would use the same syntax but simply vary behavior at runtime based on the contents of the splatted variable.
* I propose that only a single splat be allowed.

**Grammer proposal**

```
OPEN helperName param* hashSplat? hash? CLOSE
```

Visually, it makes sense to put the splat before any hash arguments. This would aid in the understanding that hash arguments should take precedence. For example given:

```handlebars
{{! presuming myHash === { keyA: 'val', keyB: context.someProp } }}
{{my-component someParam *myHash keyB="I take the crown!"}}
```

The hash would be `{ keyA: 'val', keyB: 'I take the crown!' }`. This is actually runtime behavior, and could vary according to implementation, but seems a good justification for the above ordering.

It is also suggested that subexpressions would also accept a splat. For example:

```handlebars
{{my-component someParam=(translate *translateOptions)}}
```

**Alternatives**

It has been suggested by @wycats that we consider `...myHash`, as it aligns more closely with JavaScript splatting. This seems reasonable, I'd like some more thoughts.

/cc @kpdecker @mmun @krisselden @wycats @chrislopresto

mmun:
I like this. It would be useful for Ember components. I prefer `*`. Perhaps `...` for params.

kpdecker:
Being a bum on vacation right now but quick comment:

Allowing this for hash and not array parameters feels weird and I suspect that people are going to ask for array as well. The only thing that I can think of to support that would be having a separate syntax for one vs. the other:

```
{{foo someParam ...restParams *hashParams myKey=myValue}}
```

mmun:
Some more context for the discussion: https://github.com/sebmarkbage/ecmascript-rest-spread

tl;dr: The JS syntax `{ x, y, ...z }` has been proposed and looking like its going to make it in.

kpdecker:
[...]

PR #1128, opened 2015-11-02
[FEATURE] Splat operator

BenjaminHorn:
Addresses: https://github.com/wycats/handlebars.js/issues/1050

### DIFF
--- a/lib/handlebars/compiler/compiler.js
+++ b/lib/handlebars/compiler/compiler.js
@@ -323,7 +323,8 @@ Compiler.prototype = {
 
   Hash: function(hash) {
     let pairs = hash.pairs,
-        i = 0, splat = null,
+        splat = null,
+        i = 0,
         l = pairs.length;
 
     this.opcode('pushHash');
--- a/lib/handlebars/compiler/compiler.js
+++ b/lib/handlebars/compiler/compiler.js
@@ -323,18 +323,28 @@ Compiler.prototype = {
 
   Hash: function(hash) {
     let pairs = hash.pairs,
-        i = 0,
+        i = 0, splat = null,
         l = pairs.length;
 
     this.opcode('pushHash');
-
     for (; i < l; i++) {
+
+      if (pairs[i].type === 'Splat') {
+        if (splat !== null) {
+          throw new Exception('Multiple splats are not supported yet');
+        }
+        splat = pairs[i].value;
+        continue;
+      }
       this.pushParam(pairs[i].value);
     }
     while (i--) {
+      if (pairs[i].type === 'Splat') {
+        continue;
+      }
       this.opcode('assignToHash', pairs[i].key);
     }
-    this.opcode('popHash');
+    this.opcode('popHash', splat);
   },
 
   // HELPERS
--- a/lib/handlebars/compiler/javascript-compiler.js
+++ b/lib/handlebars/compiler/javascript-compiler.js
@@ -421,7 +421,6 @@ JavaScriptCompiler.prototype = {
   // it onto the stack.
   lookupOnContext: function(parts, falsy, strict, scoped) {
     let i = 0;
-
     if (!scoped && this.options.compat && !this.lastContext) {
       // The depthed query is expected to handle the undefined logic for the root level that
       // is implemented below, so we evaluate that directly in compat mode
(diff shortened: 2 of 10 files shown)


lib/handlebars/compiler/compiler.js
```diff
@@ -323,18 +323,28 @@ Compiler.prototype = {
 
   Hash: function(hash) {
     let pairs = hash.pairs,
-        i = 0,
+        i = 0, splat = null,
```
kpdecker: nit: This should be on a separate line, probably above `i` so `i` and `l` are declared next to one another.

lib/handlebars/compiler/compiler.js
```diff
@@ -323,18 +323,28 @@ Compiler.prototype = {
...
         l = pairs.length;
 
     this.opcode('pushHash');
-
     for (; i < l; i++) {
+
+      if (pairs[i].type === 'Splat') {
+        if (splat !== null) {
+          throw new Exception('Multiple splats are not supported yet');
```
kpdecker: I think this is a fundamental issue that needs to be resolved before we can merge it. Support for multiple instances is something that will be confusing if it's not there but values can be assigned on either side.

It seems like we need to add finer tuned opcodes that only generate objects, i.e. something like:

```
pushHash
pushObject
{pushParam}
{pushParam}
assignToHash
assignToHash
popObject
splat
pushObject
{pushParam}
assignToHash
popObject
popHash
```

Which would end up generating javascript similar to the following:

```
hash: this.splat({key: value, key: value}, splat, {key: value})
```

lib/handlebars/compiler/javascript-compiler.js
```diff
@@ -550,6 +549,10 @@ JavaScriptCompiler.prototype = {
     }
 
     this.push(this.objectLiteral(hash.values));
+    if (splat) {
+      let splatIdentifier = 'depth' + splat.depth + '.' + splat.original; // or parts[0] ? what's the different
```
kpdecker: This should use one of the current lookup mechanisms, likely `lookupOnContext` combined with a `popStack` to maintain consistency in variable lookup.

spec/helpers.js
```diff
@@ -737,4 +737,82 @@ describe('helpers', function() {
...
+      var string = '{{foo **bar **baz}}';
+      shouldThrow(function() {
+        CompilerContext.compile(string);
+      }, Error);
+
+    });
+
+    it('splat shadowing', function() {
+      var string = '{{helper **splat occupation="pirate" }}';
+      var hash = {splat: {occupation: 'cannonball'}};
```
kpdecker: We should assert that `splat` doesn't change after the script executes.

kpdecker:
This generally looks very good. A few feedback points that I think we should address before getting in but the bulk of the work looks like it's done and great.

Thanks!

lib/handlebars/compiler/compiler.js
kpdecker: We would also want to do some special casing there for these cases:

1. No splat (Do not call `this.splat`)
2. Only splat (Directly assign the splat path to the hash field)
3. Splat in the first field (will want to pass an empty object to avoid mutating the splat field, i.e. `this.splat({}, rest)`

nathanhammond:
This PR should be considered a WIP, please ignore until somebody says otherwise. :)

lib/handlebars/compiler/compiler.js
nathanhammond: @kpdecker My first thought was to convert all `HashPairs` into `Splats` so that it could be done as a `merge` since duplication isn't banned with right-most precedence:
`{{foo bar="bar" bar="whoops" **bat **other}}` => `merge({bar: "bar"}, {bar: "whoops"}, {bat: "bat"}, {other: ...});`

That would make the way we think about each piece consistent, but what I'm discovering is that I don't know enough about the underlying Handlebars primitives to evaluate this approach.

Does that seem reasonable?

machty: @nathanhammond I was going to say that it seems more efficient to have the compiler do as much as possible to minimize the POJOs generated in favor of fewer POJOs with more props that we pass into merge, but I'm not actually sure if that's true (might be worth a JSPerf to see how that's optimized). 

Either way I'm nearly done wrapping up my first draft of the implementation, which leans toward minimizing POJOs.

kpdecker:
Closing in favor of #1149, since that is complete and reflects the final discussions in #1050. Thanks to @BenjaminHorn  and @nathanhammond for doing this initial work!

PR #1132, opened 2015-11-16
Update uglify-js to avoid vulnerability

plynchnlm:
The security advisory https://nodesecurity.io/advisories/48 says version 2.6.0 or later of uglify-js is needed.

The tests passed with the new version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "source-map": "^0.4.4"
   },
   "optionalDependencies": {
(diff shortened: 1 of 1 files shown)


plynchnlm:
@kpdecker Thanks!  Now that this is merged, could a new version be tagged to include it?

kpdecker:
Released in 4.0.5

PR #1148, opened 2015-12-10
Remove stringParams and trackIds mode

### LINKED ISSUE #1145: Time to kill stringParams mode?
`stringParams` mode was originally implemented to support the legacy pre-Glimmer/HTMLbars view layer in Ember.js, but since moving to Glimmer, there's no longer any need to support `stringParams` mode, so we should probably remove it from Handlebars proper...

...unless somewhere out there another prominent library that extends Handlebars also makes use of `stringParams`? Does anything come to mind? Any good reason to keep it around? It certainly is a pain to maintain.

machty:
Closes #1145


lib/handlebars/utils.js
```diff
@@ -103,6 +103,3 @@ export function blockParams(params, ids) {
   return params;
 }
 
-export function appendContextPath(contextPath, id) {
```
kpdecker: This will need to be removed from the handlebars-site docs.

kpdecker: As well as any references to either of these features in general/

lib/handlebars/helpers/each.js
```diff
@@ -30,15 +25,11 @@ export default function(instance) {
...
-
-        if (contextPath) {
-          data.contextPath = contextPath + field;
-        }
       }
 
       ret = ret + fn(context[field], {
         data: data,
-        blockParams: blockParams([context[field], field], [contextPath + field, null])
+        blockParams: blockParams([context[field], field], [field, null])
```
kpdecker: I think the `blockParams` utility method was basically in support of `trackIds` can we remove completely?

kpdecker:
Generally looks good, just the docs concerns and checking if we can remove `blockParams()`.

machty:
@kpdecker updated per your comments and submitted PR to update docs: https://github.com/wycats/handlebars-site/pull/136

PR #1149, opened 2015-12-10
Splat redux with es6-ish syntax

machty:
This PR includes and improves upon the first splat PR (#1128) based on discussions within #1050. Specifically, this PR provides (or will provide):
- ES6-ish `...`-based splatting syntax
- splatting for hashes
- splatting for positional params
- splatting subexpressions e.g. `...(wat)`

### DIFF
--- a/lib/handlebars/compiler/compiler.js
+++ b/lib/handlebars/compiler/compiler.js
@@ -321,18 +321,28 @@ Compiler.prototype = {
 
   Hash: function(hash) {
     let pairs = hash.pairs,
-        i = 0,
+        i = 0, splat = null,
         l = pairs.length;
 
     this.opcode('pushHash');
-
     for (; i < l; i++) {
+
+      if (pairs[i].type === 'Splat') {
+        if (splat !== null) {
+          throw new Exception('Multiple splats are not supported yet');
+        }
+        splat = pairs[i].value;
+        continue;
+      }
       this.pushParam(pairs[i].value);
     }
     while (i--) {
+      if (pairs[i].type === 'Splat') {
+        continue;
+      }
       this.opcode('assignToHash', pairs[i].key);
     }
-    this.opcode('popHash');
+    this.opcode('popHash', splat);
   },
 
   // HELPERS
--- a/lib/handlebars/compiler/compiler.js
+++ b/lib/handlebars/compiler/compiler.js
@@ -156,11 +156,11 @@ Compiler.prototype = {
 
   DecoratorBlock(decorator) {
     let program = decorator.program && this.compileProgram(decorator.program);
-    let params = this.setupFullMustacheParams(decorator, program, undefined),
+    let { params, splatMap } = this.setupFullMustacheParams(decorator, program, undefined),
         path = decorator.path;
 
     this.useDecorators = true;
-    this.opcode('registerDecorator', params.length, path.original);
+    this.opcode('registerDecorator', params.length, path.original, splatMap);
   },
 
   PartialStatement: function(partial) {
@@ -199,6 +199,7 @@ Compiler.prototype = {
     this.opcode('invokePartial', isDynamic, partialName, indent);
     this.opcode('append');
   },
+
   PartialBlockStatement: function(partialBlock) {
     this.PartialStatement(partialBlock);
   },
@@ -212,10 +213,10 @@ Compiler.prototype = {
       this.opcode('append');
     }
   },
+
   Decorator(decorator) {
     this.DecoratorBlock(decorator);
   },
-
 
   ContentStatement: function(content) {
     if (content.value) {
@@ -261,12 +262,14 @@ Compiler.prototype = {
   },
 
   helperSexpr: function(sexpr, program, inverse) {
-    let params = this.setupFullMustacheParams(sexpr, program, inverse),
+    let setup = this.setupFullMustacheParams(sexpr, program, inverse),
+        params = setup.params,
+        splatMap = setup.splatMap,
         path = sexpr.path,
         name = path.parts[0];
 
     if (this.options.knownHelpers[name]) {
-      this.opcode('invokeKnownHelper', params.length, name);
+      this.opcode('invokeKnownHelper', params.length, name, splatMap);
     } else if (this.options.knownHelpersOnly) {
       throw new Exception('You specified knownHelpersOnly, but used the unknown helper ' + name, sexpr);
     } else {
@@ -274,7 +277,7 @@ Compiler.prototype = {
       path.falsy = true;
 
       this.accept(path);
-      this.opcode('invokeHelper', params.length, path.original, AST.helpers.simpleId(path));
+      this.opcode('invokeHelper', params.length, path.original, AST.helpers.simpleId(path), splatMap);
     }
   },
 
@@ -321,28 +324,35 @@ Compiler.prototype = {
 
   Hash: function(hash) {
     let pairs = hash.pairs,
-        i = 0, splat = null,
+        i = 0,
         l = pairs.length;
 
     this.opcode('pushHash');
-    for (; i < l; i++) {
 
+    while (i < l) {
       if (pairs[i].type === 'Splat') {
-        if (splat !== null) {
-          throw new Exception('Multiple splats are not supported yet');
+        this.pushParam(pairs[i].value);
+        this.opcode('pushSplatHashPiece');
+        ++i;
+      } else {
+        let start = i;
+        let end = i;
+        while (end < l && pairs[end].type == 'HashPair') {
+          ++end;
         }
-        splat = pairs[i].value;
-        continue;
+
+        this.opcode('pushHashPiece');
+        for (let g = start; g < end; ++g) {
+          let pair = pairs[g];
+          this.pushParam(pair.value);
+          this.opcode('assignToHash', pair.key);
+        }
+
+        i = end;
       }
-      this.pushParam(pairs[i].value);
     }
-    while (i--) {
-      if (pairs[i].type === 'Splat') {
-        continue;
-      }
-      this.opcode('assignToHash', pairs[i].key);
-    }
-    this.opcode('popHash', splat);
+
+    this.opcode('popHash');
   },
 
   // HELPERS
@@ -394,19 +404,25 @@ Compiler.prototype = {
     }
   },
 
-  pushParams: function(params) {
-    for (let i = 0, l = params.length; i < l; i++) {
-      this.pushParam(params[i]);
-    }
-  },
-
   pushParam: function(val) {
     this.accept(val);
   },
 
   setupFullMustacheParams: function(sexpr, program, inverse, omitEmpty) {
     let params = sexpr.params;
-    this.pushParams(params);
+
+    let splatMap = [];
+    for (let i = 0, l = params.length; i < l; i++) {
+      let p = params[i];
+      this.pushParam(p);
+      if (p.splat) {
+        splatMap.push(i);
+      }
+    }
+
+    if (splatMap.length === 0) {
+      splatMap = null;
+    }
 
     this.opcode('pushProgram', program);
     this.opcode('pushProgram', inverse);
@@ -417,7 +433,7 @@ Compiler.prototype = {
       this.opcode('emptyHash', omitEmpty);
     }
 
-    return params;
+    return { params, splatMap };
   },
 
   blockParamIndex: function(name) {
--- a/lib/handlebars/compiler/javascript-compiler.js
+++ b/lib/handlebars/compiler/javascript-compiler.js
@@ -419,7 +419,6 @@ JavaScriptCompiler.prototype = {
   // it onto the stack.
   lookupOnContext: function(parts, falsy, strict, scoped) {
     let i = 0;
-
     if (!scoped && this.options.compat && !this.lastContext) {
       // The depthed query is expected to handle the undefined logic for the root level that
       // is implemented below, so we evaluate that directly in compat mode
@@ -505,11 +504,15 @@ JavaScriptCompiler.prototype = {
     }
     this.hash = {values: {}};
   },
-  popHash: function() {
+  popHash: function(splat) {
     let hash = this.hash;
     this.hash = this.hashes.pop();
 
     this.push(this.objectLiteral(hash.values));
+    if (splat) {
+      let splatIdentifier = 'depth' + splat.depth + '.' + splat.original; // or parts[0] ? what's the different
+      this.push([this.aliasable('container.splat'), '(', this.popStack(), ', ', splatIdentifier, ')']);
+    }
   },
 
   // [pushString]
@@ -930,6 +933,7 @@ JavaScriptCompiler.prototype = {
     }
 
     options.name = this.quotedString(helper);
+
     options.hash = this.popStack();
 
     let inverse = this.popStack(),
--- a/lib/handlebars/compiler/javascript-compiler.js
+++ b/lib/handlebars/compiler/javascript-compiler.js
@@ -499,20 +499,55 @@ JavaScriptCompiler.prototype = {
     this.pushStackLiteral(omitEmpty ? 'undefined' : '{}');
   },
   pushHash: function() {
-    if (this.hash) {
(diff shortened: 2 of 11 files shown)


machty:
@kpdecker @nathanhammond @BenjaminHorn this is ready for review now; just got splatting positional params and subexpressions to work.

Notes:

- I haven't figured out how to de-dupe the new grammar; I tried `SPLAT?` but it yielded grammar conflicts, even though to my eyes that's functionally equivalent to the grammar I explicitly specify
- positional param splatting works as follows: if there's no splats, there's no change to how helpers are called today; if there's at least one splat, instead doing a `.call` on the helper, we first invoke a new runtime function `splatArgs` and then `.apply` those to the helper function. 
- I didn't implement splatting the various helper method missing methods; I'm not sure it's worth the bloat?

lib/handlebars/compiler/compiler.js
```diff
@@ -156,11 +156,13 @@ Compiler.prototype = {
 
   DecoratorBlock(decorator) {
     let program = decorator.program && this.compileProgram(decorator.program);
-    let params = this.setupFullMustacheParams(decorator, program, undefined),
+    let setup = this.setupFullMustacheParams(decorator, program, undefined),
+        params = setup.params,
+        splatMap = setup.splatMap,
```
kpdecker: nit: Should we use destructuring here? (I forget what exact rules Handlebars has enabled, it was one of the earlier ES6 projects for me so it is likely a bit more conservative than need be)

spec/parser.js
```diff
@@ -215,6 +222,14 @@ describe('parser', function() {
...
       astFor('{{{{goodbyes}}}} {{{{/hellos}}}}');
     }, Error, /goodbyes doesn't match hellos/);
+
+    shouldThrow(function() {
+      astFor('{{foo ...}}');
+    }, Error, /Parse error on line 1/);
+
+    shouldThrow(function() {
+      astFor('{{foo ...=lol ...baz}}');
+    }, Error, /Parse error on line 1/);
```
kpdecker: Is `{{foo ...=lol baz}}` invalid as well?

lib/handlebars/runtime.js
```diff
@@ -85,6 +85,23 @@ export function template(templateSpec, env) {
       return typeof current === 'function' ? current.call(context) : current;
     },
 
+    splat: function() {
+      return Utils.extend.apply(null, arguments);
```
kpdecker: What is the first argument passed here? Are we ok with it being modified in all cases?

lib/handlebars/compiler/javascript-compiler.js
```diff
@@ -500,16 +499,53 @@ JavaScriptCompiler.prototype = {
...
     }
-    this.hash = {values: [], types: [], contexts: [], ids: []};
+    this.hashPieces = [];
   },
   popHash: function() {
-    let hash = this.hash;
-    this.hash = this.hashes.pop();
+    let hashPieces = this.hashPieces;
+    this.hashPieces = this.hashes.pop();
+    this.hashPiece = this.hashPieces && this.hashPieces[this.hashPieces.length - 1];
```
kpdecker: What is this bit of code doing here?

lib/handlebars/runtime.js
kpdecker: Just saw [this](https://github.com/wycats/handlebars.js/pull/1149/files#diff-12fc6be51b9642b813f72c8bd16d891aR528) so please disregard :) Do we have test coverage asserting that the first object is not mutated?

lib/handlebars/compiler/javascript-compiler.js
```diff
@@ -638,10 +674,27 @@ JavaScriptCompiler.prototype = {
...
         (helper.paramsInit ? ['),(', helper.paramsInit] : []), '),',
         '(typeof helper === ', this.aliasable('"function"'), ' ? ',
-        this.source.functionCall('helper', 'call', helper.callParams), ' : helper))'
+        this.helperFunctionCall('helper', helper, null), ' : helper))'
     ]);
   },
 
+  helperFunctionCall: function(helperName, helperOptions, splatMap) {
+    if (splatMap) {
+      let splatMapObj = {};
```
kpdecker: Any thoughts on performance/code size of this vs. array with missing values notation? I.e. `[,,1]` vs. `{"2":1}`.

lib/handlebars/compiler/javascript-compiler.js
```diff
@@ -638,10 +674,27 @@ JavaScriptCompiler.prototype = {
...
+      for (let i = 0, l = splatMap.length; i < l; ++i) {
+        splatMapObj[splatMap[i]] = 1;
+      }
+
+      let argsWithSplatMap = helperOptions.params.slice();
+      argsWithSplatMap.push(this.objectLiteral(splatMapObj));
```
kpdecker: nit: Can we do a `concat` here to avoid the intermediate array from the slice operation?

lib/handlebars/runtime.js
```diff
@@ -85,6 +85,23 @@ export function template(templateSpec, env) {
...
+    },
+
+    splatArgs: function() {
+      let splatMap = arguments[arguments.length - 1];
```
kpdecker: nit: Would the code be clearer here if we did `splatArgs(args, splatMap) {}` rather than trying to extract from the end of the args list? Alt: Lets pass `splatMap` as the first parameter so we can avoid the `.length - 1` gymnastics.

lib/handlebars/compiler/javascript-compiler.js
machty: What use to just be stack of hash objects is now a stack of hash piece arrays, and `this.hashPiece` refers to the last (current) hashPiece, so we restore it to that last element unless we've popped off all the hash piece arrays.

kpdecker:
@machty Generally this looks really good, had a few minor comments. Responding to your notes:
1) I've seen similar issues when trying to do the same. Sometimes Jison gives ambiguous productions for things that to me seem like they should be unambiguous and are the same as the duplicated listing. Since it's in the compiler I'm not too worried about it not being completely DRY.
2) Perfect
3) For blockHelperMissing, that code is only run if the template has no args or hash, so you are correct that no changes are needed.

spec/parser.js
machty: Yes, I'll add a test. Basically this PR doesn't change the rule that params are on the left, hashy things on the right.

lib/handlebars/compiler/javascript-compiler.js
kpdecker: Ok, maybe we can put a comment in the code to explain the restore operation?

kpdecker:
Pong?

machty:
We've been holding off on pulling the trigger on this while we flesh out some remaining semantics on the Ember side of things, particularly related to how splatting will work for so-called angle-bracket components.

joeyjiron06:
when do you think this will get in? im desperately waiting for this :)

rtablada:
@machty I want http://s3-ak.buzzfed.com/static/2014-07/10/20/enhanced/webdr10/anigif_enhanced-16641-1405037819-19.gif

morhook:
I like this!!

foxnewsnetwork:
While we're waiting spread and splat to come out in 5.0.0 handlebars, I wrote up a small Ember addon that handles spreading for Ember's `component` helper here: https://github.com/foxnewsnetwork/ember-component-helpers

That can be used by users of Ember2.0 and above like this:
```javascript
Ember.Controller.extend({
  arrayArgs: ["admin.posts.post", 33, {class: "mdl-navigation__link", tagName: "li"}]
});
```
```handlebars
{{#component-apply "link-to" arrayArgs}}
  Your Link Text
{{/component-apply}}
```

You know, in case anyone else is particularly hungry for the spread after 1 year of waiting for it to come out like 2 straight months of looking at @rtablada 's animated gif of butter being spread on bread

workmanw:
Just curious if there is any update on this. We could badly use this in our app. :)

vomchik:
+1

knownasilya:
What's the status on this?

sglanzer:
For anyone looking for another option while waiting https://github.com/ciena-blueplanet/ember-spread is mixin based, but works with dynamic hashes

knownasilya:
On another note, splats might work better as Element Modifiers (https://github.com/emberjs/rfcs/pull/112), e.g. `{{my-comp (splat params)}}` or `<my-comp {{splat params}}/>`

lib/handlebars/compiler/javascript-compiler.js
jamesarosen: `push` mutates. `concat` returns a new array. Either way the goal seems to be to create a new array to avoid mutating `helperOptions.params`. Though I suppose VMs might optimize `a.concat(b)` in ways they can't `a.slice().push(b)`.

GCheung55:
Any updates on this? How can people help to get this merged?

gnapse:
Still nothing on this? I keep coming back to needing this about once a month. And every time I have to make compromises in my architecture design because of this. If there's anything on which help is needed, I'd be happy to try and help.

machty:
I'd like to get a sense of how many people would use this feature/syntax for splatting arrays, spreading a hash of options. or both? This may or may not impact the future of this feature, I'm mostly just curious as to the real world use cases people expect this to address.

Personally, 99% of the time I'd be using this to spread a hash of options, since I do a lot of work in Ember with components and it's a common desire to be able to create a wrapper component that might introduce additional KV pairs into the options hash but otherwise forward the hash to the wrapped component. I don't think I'd ever use array splatting.

pjcarly:
I'd use it exactly as you describe @machty 

I currently have situations with a wrapper component like this, but then with 5 or 6 attributes that are being passed.

_index/template.hbs_
`{{wrapper-component arg1=arg1 arg2=arg2 arg3=arg3}}`

_wrapper-component/template.hbs_
`{{wrapped-component arg1=arg1 arg2=arg2 arg3=arg3}}`

it would be nice to not have to duplicate the passing of arguments in the wrapper-component's template

workmanw:
👍  on a hash of options (named args). Probably wouldn't use the array much.

foxnewsnetwork:
@machty From my personal experience, this is mostly an issue when trying to build a component to wrap `{{link-to}}` components (as oppose to extending `LinkComponent`). For example, if I were to create a `{{my-fancy-link-to}}` component that looks something like:

`app/components/templates/my-fancy-link-to.hbs`
```hbs
{{yield}}
{{#link-to ...args }}
  <div class='fancy-stuff'>...</div>
{{/link-to}}
<div class='other-links'>
  {{#link-to "somewhere.else"}}
    ...
  {{/link-to}}
</div>
```

I would need to "splat" the `...args` to properly emulate the functionality, but at the same time, I have no easy way of doing the splat

gnapse:
I too have always looked for something like this for spreading a hash of options. And I also wanted to know that it's not just to avoid duplicating the enumeration of options in the component and the wrapper. It's also to make the wrapper component agnostic of the component it is wrapping. I've had use cases with components that could conceivably wrap totally different inner components, with largely different sets of possible options. The wrapper component need not know about these kind of stuff.

XaserAcheron:
Confession time: when I was first learning Ember, I resorted to some shameful hacking to get around the exact case of having to pass a zillion variables through a wrapper component. I eventually gave in and started doing things the proper way, but being able to splat the hash (eww) would really save both my past and present self a few brain cells.

LevelbossMike:
I'd also like to see this for the use-case of wrapper components. Right now I'm using ember-spread to do this:

http://blog.firstiwaslike.com/clean-ember-addon-component-customization-with-ember-spread/

but it would be great if that was built-in functionality.

jamesarosen:
I have another potential use-case for this. I have a pagination-controls component that builds up a list of pages, then does
```hbs
{{#each pages as |page|}}
  {{#link-to (query-params page=page.number)}}page.number{{/link-to}}
{{/each}}
```
I'd like to make the name of the query-param configurable:
```hbs
{{my-pagination param='foosPage'}}
```

I can think of two ways of doing that, both of which require new syntax

```hbs
// interpolate hash key:
{{#link-to (query-params [param]=page.number)}}

// hash-splat:
{{#link-to (query-params ...page.queryParamsHash)}}
```

knownasilya:
Also the array splat is pretty important for positional params, `{{#link-to dynamicRouteName ...routeParams}}`. Currently, you cannot use `link-to` with dynamic params.

gabrielgrant:
Is this only going to work for components, or will this also work for passing params to helpers? In practice, helpers seem to take positional params far more frequently than components do

ldong:
Cannot wait to see this get merged.

stevenvachon:
I thought it was called "spread" in ES2015.

jbescoyez:
@machty Thanks for this awesome PR. I'm wondering how comes this has not been merged yet. 

3 questions:
* Is there some blocker at the moment?
* What workarounds are you using for handling dynamic options passing between component?
* How can I help to see this merged?

lib/handlebars/compiler/javascript-compiler.js
machty: I don't know, don't have the perf wizard knowledge myself.

machty:
Just pushed a rebase to fix merge conflicts.

@jbescoyez 

1. Basically the issue is that a large enough threshold of people are wanting this for Ember, but even if we land this feature, the implementation for template data-binding on the Ember/Glimmer side of things still needs to happen, and particularly with Glimmer there are performance constraints that need to be considered before we ship a syntax that everyone might like but is unfortunately doomed to be untenably slow.
2. See these comments from this thread: https://github.com/wycats/handlebars.js/pull/1149#issuecomment-217692299 and https://github.com/ciena-blueplanet/ember-spread
3. I'm not sure, it's really mostly blocked on Glimmer, which is developing quite rapidly (though admittedly this feature's been on the backburner). I've pinged the Ember/Glimmer team as to how to proceed with this PR. I'm thinking though that it probably makes sense for there to be an official Ember/Glimmer RFC with some backing from the core team, like we did with Named Blocks.

nknapp:
Basically @wycats has to say that it should get merged. I promised only to fix bug and not add new features, when I asked for push access.

He wants a spec first, have a look at #1277

NindroidX:
Are there any plans on merging this (or implementing something similar)? Any news?

lsg-braymon:
Any updates or news on this PR? Thanks!

nknapp:
There is no update and for the last 2 years, no one has made a serious attempt to write a spec. I'm thinking about it from time to time, but it's a lot of work, I have different obligations and for me, it would be just for the fun of it. I have actually started an attempt [here](https://github.com/handlebars-lang/specification) but it has been unchanged for a long time.

I doubt that anything will change here unless somebody (other than me) gets invested in this an starts writing.

nknapp:
Hi @machty. This PR is now open for 4 years and I feel pity that you have put ao much work in a PR that has then been ignored for so long. I haven't really looked at it until now because it was before my time, and @wycats asked me to only fix bugs while this is clearly a feature. That hasn't changed in principal and after talking to @wycats I think that development will go into another direction.

But I can ask @wycats to have a look and give his approval for merging this. **I will do this only if you say you still need it.**

The reason why I'm asking this is: I want to change some tooling and use "prettier" as a code formatter. This will make development much easier in the future, but it will make merging this request much more difficult.

I could just discard it, but I don't want to do that without asking first.

knownasilya:
Is love to see this make it in.

machty:
I'm not going to be bothered at all if this ends up getting closed. Even if it does, it'll still leave a blueprint for what needs to happen to get this over the finish line.

I don't have the bandwidth to be a champion of this feature, but I believe one of the main blockers was that it wasn't clear whether implementing a "splat/spread" operator was going to be an easily-optimizable feature in Glimmer. But regardless of efficiency, it seems like an obvious win for component composition (and proven by React and others) to be able to forward all the args to a child component without enumerating all of them. 

So I'd say close it for now so you can evolve the codebase and start using Prettier, but I wouldn't expect the debate to stop.

nknapp:
I have some concerns that templates using this feature will become more difficult to understand, when trying to track the current evaluation context.

I'll gladly close it for now. I just didn't want to upset you.

PR #1153, opened 2015-12-14
Adding documentation for running tests

ryanmurakami:
Not sure if there's a different way to get the mustache spec dependency. This is how I did it, so let me know if there's a better way.


kpdecker:
Glad to accept this, but it should go into CONTRIBUTING.md

ryanmurakami:
Got it! Let me know if there's anything else.

PR #1176, opened 2016-02-05
Move uglify from optionalDependencies to devDependencies

mourner:
`optionalDependencies` causes `uglify-js` to be fetched every time NPM installs `handlebars` or any modules that depend on it, which is wasted bandwidth (2MB). It should be in `devDependencies` instead. cc @kpdecker


kpdecker:
This is used by end users who use the CLI interface.

PR #1177, opened 2016-02-05
Preserve License info in Closure Compiler

gennadiylitvinyuk:
To preserve license info in Closure Compiler the license has to be JSDoc-comment (not simple comment) and have @license before license text


kpdecker:
No one is going to come after you if this isn't in the minimized output, but merging anyway.

PR #1182, opened 2016-02-16
Update lookup.js

superhero:
Makes it possible to use lookup with multiple arguments, eg: `{{lookup obj 'foo' 'bar' 'baz'}}` instead of `{{lookup (lookup (lookup obj 'foo') 'bar') 'baz'}}`

### DIFF
--- a/lib/handlebars/helpers/lookup.js
+++ b/lib/handlebars/helpers/lookup.js
@@ -1,8 +1,13 @@
 export default function(instance) {
   instance.registerHelper('lookup', function(obj) {
     var args = Array.prototype.slice.call(arguments, 1);
-    while (args.length > 1) {
-      obj = obj[args.shift()];
+    try {
+      while (args.length > 1) {
+        obj = obj[args.shift()];
+      }
+    }
+    catch(e) {
+      return undefined;
     }
     return obj;
   });
--- a/lib/handlebars/helpers/lookup.js
+++ b/lib/handlebars/helpers/lookup.js
@@ -1,5 +1,9 @@
 export default function(instance) {
-  instance.registerHelper('lookup', function(obj, field) {
-    return obj && obj[field];
+  instance.registerHelper('lookup', function(obj) {
+    var args = Array.prototype.slice.call(arguments, 1);
+    while (args.length > 1) {
+      obj = obj[args.shift()];
+    }
+    return obj;
   });
 }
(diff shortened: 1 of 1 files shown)


kpdecker:
`lookup` is intended for use with variables names. I can imagine use cases where there are a lot of nested variably named fields, but it seems uncommon enough that we shouldn't penalize (WRT performance) the normal use case. Do you have any code examples that show the problem that you are trying to solve in your own code?

erik-landvall:
@kpdecker If there is any code example; there is an example in the comment above, idk how detailed example you looking for. You say that this will penalize performance, this marginal difference seems to small to mind, imo, for the benefit it offers. How ever, if so is the case, we could implement it as a new helper; `lookupPath`. Personally, it don't matter to me ofc, I just registered the helper manually.

kpdecker:
`lookup` doesn't make sense with static keys though, which is where my confusion comes from. `(lookup (lookup var2) var1)` is the only reason that you should have to nest these and that doesn't feel like a good data model design to be honest.

erik-landvall:
If you like to reject he proposition, I won't be offended :) the use case is for what's stated by us both. Weather nested objects are good design or not, is beside the discussion imo.

kpdecker:
The distinction is variable names. You shouldn't have to pass in a constant
string to lookup since you can just do foo.bar and that's the part that I'm
missing.

erik-landvall:
Yes but you can't do `foo['bar'].baz` if you must "look up" the first child, it's children also must be "looked up". Imo, the preferred and expected behavior would be `(lookup foo 'bar').baz`

erik-landvall:
...also you can't do `foo['bar']['baz']` but it should go with out saying :)

erik-landvall:
`foo.bar.baz` will only work if you know the value of the string, understand that the strings are variables, or at least the lookup that needs to take place.. so in JS `foo[bar][baz]` can't be done in handlebars with out nested lookups.

kpdecker:
Ok, and there lies the problem. I agree that `(lookup foo 'bar').baz` should be possible and it's worth investigating adding that to the language.

erik-landvall:
It would still cause issues with the last example though: `foo[bar][baz]` where you need to lookup deeper. The prefererad way to solve it is to not have to use a helper for this at all, but to implement something like the vanilla JS way, where you can just simply type: `foo[bar][baz]`. But until then, I say that this helper will do until that issue has been addressed :) I mean, the helper is just here to complete what should be in the syntax from the beginning. This helper completes the expected behavior better.. again, at least imo :)

nknapp:
I'm closing this, because the discussion is 2 years old and I believe that the need is gone (or has been worked around).

PR #1187, opened 2016-02-19
Ensure that existing blockParams and depths are respected on dupe programs

### LINKED ISSUE #1186: @partial-block interacting strangely with identical helper blocks.
The minimal test case:
```
{{#*inline "test"}}
  <div>{{> @partial-block }}</div>
{{/inline}}

{{#>test }}
  {{#each listOne as |item|}}
    <p>{{ item }}</p>
  {{/each}}
{{/test}}

{{#>test }}
  {{#each listTwo as |item|}}
    <p>{{ item }}</p>
  {{/each}}
{{/test}}
```

With this input:
```
{ 
  listOne: ["abc", "def"],
  listTwo: ["123", "zxy"]
}
```

Produces this error and stack-trace under Handlebars 4.0.5 (also a confirmed bug via the 4.0.3 tryhandlebarsjs.com REPL):
```
TypeError: Cannot read property '0' of undefined
    at eval (eval at createFunctionContext (…/node_modules/handlebars/dist/cjs/handlebars/compiler/javascript-compiler.js:254:27), <anonymous>:8:167)
    at prog (…/node_modules/handlebars/dist/cjs/handlebars/runtime.js:224:12)
    at execIteration (…/node_modules/handlebars/dist/cjs/handlebars/helpers/each.js:51:19)
    at Object.<anonymous> (…/node_modules/handlebars/dist/cjs/handlebars/helpers/each.js:61:13)
    at eval (eval at createFunctionContext (…/node_modules/handlebars/dist/cjs/handlebars/compiler/javascript-compiler.js:261:27), <anonymous>:14:31)
    at prog (…/node_modules/handlebars/dist/cjs/handlebars/runtime.js:224:12)
    at Object.invokePartial (…/node_modules/handlebars/dist/cjs/handlebars/runtime.js:273:12)
    at Object.invokePartialWrapper [as invokePartial] (…/node_modules/handlebars/dist/cjs/handlebars/runtime.js:68:39)
    at Object.eval (eval at createFunctionContext (…/node_modules/handlebars/dist/cjs/handlebars/compiler/javascript-compiler.js:261:27), <anonymous>:6:28)
    at main (…/node_modules/handlebars/dist/cjs/handlebars/runtime.js:173:32)
```

I *think* the summary for this bug is:
* The last block,
* … in a run of identical helper blocks (eg. the contents in between the {{#each ...}} {{/each}} pairs)
* … that occur within a `@partial-block`-using wrapper partial
* … will explode.

(To be clear, the example I hit this with does *not* use in-line partials; I'm only using them here to get a smaller example. The error occurs both with and without the use of them.)

charleso:
Fixes #1186 

I'm opening this partly just to help identify the problem, my knowledge of the internals of `handlebars.js` is non-existent and there may well be more desirable ways to fix this problem.

I believe underlying cause is that the previous `compiler.useBlockParams` is not respected for any subsequent existing programs that are matched. I've just set/stashed/slashed/hacked the value on to `child` environment, but this is the part that may need to be done differently.

Question - how much performance does matching/re-using the existing programs save? (Just curious)

### DIFF
--- a/lib/handlebars/compiler/javascript-compiler.js
+++ b/lib/handlebars/compiler/javascript-compiler.js
@@ -701,11 +701,11 @@ JavaScriptCompiler.prototype = {
       child = children[i];
       compiler = new this.compiler();    // eslint-disable-line new-cap
 
-      let index = this.matchExistingProgram(child);
+      let existing = this.matchExistingProgram(child);
 
-      if (index == null) {
+      if (existing == null) {
         this.context.programs.push('');     // Placeholder to prevent name conflicts for nested children
-        index = this.context.programs.length;
+        let index = this.context.programs.length;
         child.index = index;
         child.name = 'program' + index;
         this.context.programs[index] = compiler.compile(child, options, this.context, !this.precompile);
@@ -714,12 +714,14 @@ JavaScriptCompiler.prototype = {
 
         this.useDepths = this.useDepths || compiler.useDepths;
         this.useBlockParams = this.useBlockParams || compiler.useBlockParams;
+        child.useDepths = this.useDepths;
+        child.useBlockParams = this.useBlockParams;
       } else {
-        child.index = index;
-        child.name = 'program' + index;
+        child.index = existing.index;
+        child.name = 'program' + existing.index;
 
-        this.useDepths = this.useDepths || child.useDepths;
-        this.useBlockParams = this.useBlockParams || child.useBlockParams;
+        this.useDepths = this.useDepths || existing.useDepths;
+        this.useBlockParams = this.useBlockParams || existing.useBlockParams;
       }
     }
   },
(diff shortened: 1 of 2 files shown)


lib/handlebars/compiler/javascript-compiler.js
```diff
@@ -714,12 +714,14 @@ JavaScriptCompiler.prototype = {
 
         this.useDepths = this.useDepths || compiler.useDepths;
         this.useBlockParams = this.useBlockParams || compiler.useBlockParams;
+        child.useDepths = this.useDepths || compiler.useDepths;
+        child.useBlockParams = this.useBlockParams || compiler.useBlockParams;
```
kpdecker: We don't need the `||` conditions here, it's already been done in the lines above.

kpdecker:
Thanks. One minor nit and otherwise this looks great.

charleso:
@kpdecker Good point. Updated.

PR #1217, opened 2016-05-04
Fix Comparsion Operation to Strict Equal

kazikai:
The "==" or "!=" operator is not  strict comparison operator.
It means equality without type coercion.
so we  should be avoided.


kazikai:
Hi, Recently I commit to Handlebars Github Proejct 
But. my commit is failed in CI Build. Why do I build fails?
Thank You,

stevenvachon:
Try changing the `=== null` back to `== null` as this will catch both `null` and `undefined`.

kazikai:
Thank You, your response!: I understand it

nknapp:
The failing test seems to be related to PR #1180 and not this one.

PR #1269, opened 2016-11-09
Eslint upgrade and fixes

travnels:
Upgrading grunt-eslint from 17.1.0 -> 19.0.0 resulted in the following:
- Identified duplicate rules in the .eslintrc file
- Identified rules that have been replaced or needed slight modification 
- Identified a variable that was assigned but never used - depthString in lib/handlebars/compiler/helpers.js

Added node v6 to travis-ci since it is now active LTS (https://github.com/nodejs/LTS)


marado:
Works for me.

.eslintrc
```diff
@@ -43,7 +43,6 @@
     "no-empty-character-class": 2,
     "no-ex-assign": 2,
     "no-extra-boolean-cast": 2,
-    "no-extra-parens": 0,
```
nknapp: I'd like to know what's exactly duplicate here? Is 0 the default value?

.eslintrc
```diff
@@ -179,7 +178,6 @@
     "one-var": 0,
     "operator-assignment": 0,
     "padded-blocks": 0,
-    "quote-props": 0,
```
nknapp: I'd like to know what's exactly duplicate here? Is 0 the default value?

.eslintrc
travnels: no-extra-parens is found twice in that file. `"no-extra-parens": 0` and `"no-extra-parens": [2, "functions"]`. The upgrade of grunt-eslint from 17.1.0 -> 19.0.0 identified this.

.eslintrc
travnels: no-extra-parens is found twice in that file. `"quote-props": 0` and `    "quote-props": [2, "as-needed", {"keywords": true}]`. The upgrade of grunt-eslint from 17.1.0 -> 19.0.0 identified this.

nknapp:
It would be great if you could provide a link to the documentation where you got those rule changes from. I don't know what's the difference between `keyword-spacing: 0` and `keyword-spacing: 2` for example.

I would like to merge the PR. It does not seem to be controversial. 
But I would like to understand the changes before I do that.

I would probably extract the last commit into a separate PR and squash the other ones. Is that OK for you?

nknapp:
I won't do it know. Friday is a good day.

travnels:
Removing the duplicate rules is required with the upgrade to grunt-eslint 19.x. Since there were duplicated in the file, I removed the one that was shutting off the rule (`no-extra-parens: 0` and `quote-props: 0`). This can be found in the docs: http://eslint.org/docs/user-guide/configuring#configuring-rules - `"off" or 0 - turn the rule off`. I think there can be a lot of cleanup in this file. It might be good to explore the use of current defaults and add the rules that are required for the current code base to not have errors.

nknapp:
Ah, thanks. The part with `0=off, 1=warn, 2=error` was the one that I was missing. I'll add some more comments to the commits.

.eslintrc
```diff
@@ -55,7 +54,7 @@
     "no-negated-in-lhs": 2,
     "no-obj-calls": 2,
     "no-regex-spaces": 2,
-    "quote-props": [2, "as-needed", {"keywords": true}],
+    "quote-props": [2, "as-needed", {"keywords": true, "unnecessary": false}],
```
nknapp: Ok, I think this is the last question I have so far. { "unnecessary": false } seems to be a relaxation of the current rules. Was it the default in the previous version? Did you have compile errors, without this option?

travnels: Without that, the current code base throws 40 "Unnecessarily quoted property" errors.
For example, in regressions.js:

    4:7   error  Unnecessarily quoted property 'books' found        quote-props
    5:9   error  Unnecessarily quoted property 'title' found        quote-props
    6:9   error  Unnecessarily quoted property 'author' found       quote-props
    7:11  error  Unnecessarily quoted property 'name' found         quote-props

nknapp:
This PR is merge as #1294 without the Travis-CI configuration update.

@travnels I have used the "squash and merge"-button to merge the PR and actually expected you to remain the author of the commit, but it looks like the github put me in as the author now. I'm sorry that I "stole" your contribution. I referenced you in the commit message, but it is still sad. But I can't undo it anymore without a force-push to master and that should never be done.

PR #1272, opened 2016-11-12
Fix istanbul dev dependency

marado:
According to npm's documentation[1], dependencies on github
repositories are declared with the following syntax:
    "module": "person/module"

This commit changes the package.json syntax on istanbul's dependency,
to follow the documentation.

[1] https://docs.npmjs.com/files/package.json#github-urls

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "grunt-eslint": "^17.1.0",
     "grunt-saucelabs": "8.x",
     "grunt-webpack": "^1.0.8",
-    "istanbul": "github:kpdecker/istanbul",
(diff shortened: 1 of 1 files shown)


travnels:
Shouldn't we be using the NPM package (https://www.npmjs.com/package/istanbul), instead of the git repo?

marado:
This fork introduces something that's being used on handlebars.js and
still(?) not merged upstream.

nknapp:
Mhm, you are right. The "github:user/repo"-syntax is not mentioned in the documentation, although it seems to work.

Issue #1273, opened 2016-11-14
be helpful to LibreJS


marado:
LibreJS is a set of tools that inspect website's javascript code to find out if that code is free software licensed or not.

Helping LibreJS realize handlebars.js is free software requires a small change on how the end result files are presented, as you can see here:

https://github.com/marado/kokori/commit/8a4a39b2c6501c60a76a0b71c13779fb48b602ee#diff-0492b7185eb2509fd87c513623b3ef7c

This seems a trivial change to handlebars.js, without any kind of impact, so I'm hoping it will be done on this project.

Thanks

PR #1274, opened 2016-11-14
add more machine-readable data about the license

marado:
This small change is meant to make LibreJS happy.

Closes: #1273

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -22,8 +22,9 @@ module.exports = function(grunt) {
       dist: {
         options: {
           processContent: function(content) {
-            return grunt.template.process('/**!\n\n @license\n <%= pkg.name %> v<%= pkg.version %>\n\n<%= grunt.file.read("LICENSE") %>\n*/\n')
-                + content;
+            return grunt.template.process('/**!\n\n @license magnet:?xt=urn:btih:d3d9a9a6595521f9666a5e94cc830dab83b65699&dn=expat.txt Expat\n <%= pkg.name %> v<%= pkg.version %>\n\n<%= grunt.file.read("LICENSE") %>\n*/\n')
+                + content
+		+ '\n// @license-end\n';
           }
         },
         files: [


Gruntfile.js
```diff
@@ -22,8 +22,9 @@ module.exports = function(grunt) {
       dist: {
         options: {
           processContent: function(content) {
-            return grunt.template.process('/**!\n\n @license\n <%= pkg.name %> v<%= pkg.version %>\n\n<%= grunt.file.read("LICENSE") %>\n*/\n')
-                + content;
+            return grunt.template.process('/**!\n\n @license magnet:?xt=urn:btih:d3d9a9a6595521f9666a5e94cc830dab83b65699&dn=expat.txt Expat\n <%= pkg.name %> v<%= pkg.version %>\n\n<%= grunt.file.read("LICENSE") %>\n*/\n')
```
lawnsea: Where does this magical value come from?

marado:
Oops, yes, I should have said that...

It comes from https://www.gnu.org/software/librejs/free-your-javascript.html
.

lawnsea:
@marado Ok, thanks. Do we also need `/* @license-end */`?

marado:
That's how I read the documentation, yes.

nknapp:
I just was a bit confused, because there is no explicit magnet-link for the MIT-license on the page [on the libreJS-page](https://www.gnu.org/software/librejs/free-your-javascript.html). It just says:

> Please note: When people speak of the "MIT license" they mean either the X11 license or the expat license. Please see which license the code uses, and label it accordingly.

But the [expat-license](http://directory.fsf.org/wiki/License:Expat) is what's in Handlebars' LICENSE.md file, so everything is fine.

I would like to merge the PR. I would normally do 'squash and merge' or 'rebase and merge' but that would make me the author of the commit (I'm not sure with 'rebase and merge'). Could you rebase the branch onto of master so that I can merge it?

marado:
Since this is only one commit, there's nothing really to squash, and rebase and merge *should not* create a new commit (thus not crunching the authorship of the patch). So feel free to go ahead and 'rebase and merge'. If in any case github ends up munching the authorship, well, I won't loose any sleep over it either :-)

marado:
BTW, the magnet links *are* on that page, check section 3.2.2.1 (List of magnet links for valid licenses).

nknapp:
You're right. It's still yours.

Yes, the magnet links *are* on the page, yes, but there is no link for MIT, just for "expat" and "X11".
I just read "expat"-license for the first time, so I wanted to look.

marado:
Oh, I see.
"MIT License" doesn't really refer to one particular license: the term is used for several (slightly different) licenses, so the term 'expat' is used for the particular variant used in Handlebars.
More info, if you're curious: https://en.wikipedia.org/wiki/MIT_License#Variants

PR #1282, opened 2016-12-02
Template for issues and pull-request

nknapp:
I have noticed that many issues are initially answered with "Could you please provide a jsfiddle to reproduce your problem". Since GitHub now supports [templates for issues and pull-requests](https://github.com/blog/2111-issue-and-pull-request-templates), I thought it might be a good idea to create a template that asks for a jsfiddle from the start and points to the template directly.
This saves the time for collaborators and probably leads to a faster solution for the reporter.

I have provided a pull-request-template as well and it is mostly copied from the contributing-guide.


nknapp:
If those texts are good and nobody disagrees in that such template are useful, I would like to merge the PR (and will probably do so tomorrow).

Issue #1284, opened 2016-12-19
`column` property from error messages is non-enumerable in 4.0.6 with `noEscape: true` (breaking regression)


nathanboktae:
We had a test that tried to compile this simple template and verify the error result:

```handlebars
SELECT sum(gold_balance) FROM {{#if}}_p_user__gold_balance{{/foo}}
```

```json
{
  "column": 33,
  "lineNumber": 1,
  "message": "if doesn't match foo - 1:33",
  "name": "Error"
}
```

but it started failing on `4.0.6`. Rolling back to `4.0.5` solves it.

```
~/work/api-server 23bc2a8|merge-dashboard-queries  ✹
 9:12AM 1 ᐅ npm install
api-server@0.0.1 /Users/nblack/work/api-server
└─┬ handlebars@4.0.6
  └── uglify-js@2.7.5

~/work/api-server 23bc2a8|merge-dashboard-queries  ✹
 9:13AM ᐅ TZ=Etc/UTC mocha -g "templated query fails"

  Queries
    Dashboard Queries
      1) should return 400 if the templated query fails to compile


  0 passing (254ms)
  1 failing

  1) Queries Dashboard Queries should return 400 if the templated query fails to compile:

      Error: expected { errors:
   { query:
      { column: 33,
        lineNumber: 1,
        message: 'if doesn\'t match foo - 1:33',
        name: 'Error' } } } response body, got { errors:
   { query:
      { lineNumber: 1,
        message: 'if doesn\'t match foo - 1:33',
        name: 'Error' } } }
      + expected - actual

       {
         "errors": {
           "query": {
      +      "column": 33
             "lineNumber": 1
             "message": "if doesn't match foo - 1:33"
             "name": "Error"
           }

      at endReadableNT (_stream_readable.js:926:12)
      at _combinedTickCallback (internal/process/next_tick.js:74:11)
      at process._tickDomainCallback (internal/process/next_tick.js:122:9)




~/work/api-server 23bc2a8|merge-dashboard-queries  ✹
 9:13AM 1 ᐅ npm install
api-server@0.0.1 /Users/nblack/work/api-server
└── handlebars@4.0.5

~/work/api-server 23bc2a8|merge-dashboard-queries  ✹
 9:14AM ᐅ TZ=Etc/UTC mocha -g "templated query fails"

  Queries
    Dashboard Queries
      ✓ should return 400 if the templated query fails to compile


  1 passing (212ms)
```

rafde:
[...]

PR #1285, opened 2016-12-20
Make "column"-property of Errors enumerable

nknapp:
This PR contains a test case for #1284. It verifies that the column-property is included in the compilation error. The test passes. This is not mean to be merged right now, primarily for discussion.

### DIFF
--- a/lib/handlebars/exception.js
+++ b/lib/handlebars/exception.js
@@ -31,7 +31,10 @@ function Exception(message, node) {
       // Work around issue under safari where we can't directly set the column value
       /* istanbul ignore next */
       if (Object.defineProperty) {
-        Object.defineProperty(this, 'column', {value: column});
+        Object.defineProperty(this, 'column', {
+          value: column,
+          enumerable: true
+        });
       } else {
         this.column = column;
       }
--- a/spec/compiler.js
+++ b/spec/compiler.js
@@ -54,6 +54,15 @@ describe('compiler', function() {
       }
     });
 
+    it('should include the location as enumerable property', function() {
+      try {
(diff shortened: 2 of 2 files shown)


nknapp:
I have reworded my commit messages and would not consider this work-in-progress anymore. If all tests pass and this PR does not introduce another regression (I have no Safari to test the bug fixed in 20c965c), it could be merged (from my point of view) and it would probably help @nathanboktae a lot if there was a 4.0.7 release including this PR.

@kpdecker @lawnsea @ErisDS @wycats

nathanboktae:
> I have no Safari to test

That's what SauceLabs is for right? I'm not seeing right off hand if Travis CI is kicking off tests in SauceLabs

nknapp:
@wycats gave me commit access tonight. I have created a local branch "tmp/stage-4.0.7" and push the changes there. This should run the tests in SauceLabs and show any Safari-problems.

nknapp:
Tests case failing for Safari 8: ['97' should === '5': Checking error column"](https://github.com/wycats/handlebars.js/pull/1285/commits/7a51fa7cb24bc77b2e9c555619d34f6c11a50c94#diff-49874fc4bd5a81e1ceb0e8d284ee8c8bR46).

Why should Safari 8 return a completely different column than all other browsers?

nknapp:
Ah, that's why: https://github.com/jquery/esprima/issues/1290#issuecomment-132455482

nknapp:
The column-property is read-only in Safari 8, its value won't change, even with Object.defineProperty. The test case now omits the column-check, if the property is read-only.

PR #1289, opened 2016-12-27
Stringify the obj when using 'strict' mode

AbnerZheng:
when using 'strict' mode, message just got a [object Object], rather than a stringify json.

### DIFF
--- a/lib/handlebars/runtime.js
+++ b/lib/handlebars/runtime.js
@@ -69,7 +69,7 @@ export function template(templateSpec, env) {
   let container = {
     strict: function(obj, name) {
       if (!(name in obj)) {
-        throw new Exception('"' + name + '" not defined in ' + obj);
+        throw new Exception('"' + name + '" not defined in ' + JSON.stringify(obj));
       }
       return obj[name];
     },


nknapp:
I agree that [object Object] is not very helpful in this place and it will probably be [object Object] in most cases. Stringifying the whole object might lead to very large error messages. Maybe the Object.keys array would be enough?

AbnerZheng:
@nknapp I agree with you,  it's just a  fixed.
I'm in a situation where handlebars should warns these fields that can't be found in data but continue compiling. But I found that handlebars couldn't handle it. The only method I can figure out is to repeat compiling in strict mode and replace the error field with a constant. Below is the algorithm:
```javascript
var flag = true;
while(flag)
try{
  handlebars.compile(content,{strict: true})(data);
  flag = false;
}catch(e){
  var field = parse(e) //to get the filed that can't find in data
   console.log(field);
   content = content.replace(field, "");
}
```

But since the error message has not detail messages, this method can't work.

In  [`lib/handlebars/runtime.js#line71`](https://github.com/wycats/handlebars.js/blob/master/lib/handlebars/runtime.js#L71), the error message like below:
```javascript 
if (!(name in obj)) {
        throw new Exception('"' + name + '" not defined in ' + obj);
 }
```
There is two problems. 

1. obj is stringify to [object Object], as told above.
2. name has lose the context. For example, when data has this form: { a: { b: 1}},  and the content to parse is "{{a.c}}" then, the name will be c rather than a.c, which losing the context.

So, do you think if there any need to pass through context when recursive?

nknapp:
Yes, it's the default behavior that Handlebars ignores missing fields. 

If there are specific fields that you want to check and you want to display a constant value in your output, you can write your own helper. Something like this: https://jsfiddle.net/9D88g/58/

There is no way of getting the name of the field in this case though. You'd have to pass that as an additional parameter.

nknapp:
Some more thoughts, to illustrate that his is more complicate to decide on:

* If you say that you want to parse the error message, then you probably expect the message-format to be stable in the future. Other developers might already depend on the current format, so **this must be treated as a feature request and breaking change.** 
* The current approach is actually more flexible, because you can provide a custom toString()-method to your object (see https://jsfiddle.net/9D88g/61/). This would even allow you to implement use `JSON.stringify()` without any changes in Handlebars. So **this PR is actually not needed to solve your problem**. This solution just came to my mind while I was writing this message, but I think it would be the best option for you.
* And then there is the already mentioned concern that the string might grow very large.

That's why I would like to close the PR without merging, but I'm leaving it open for other people to see for a while.

AbnerZheng:
It works well.  
Thanks a lot for your solution.

nknapp:
That's great. I'm closing this then.

Issue #1252, opened 2016-09-21
Using @partial-block twice in a template not possible


Dangoo:
I discovered that my issue with #1185 is caused by including the `@partial-block` twice inside a template like this:

**template.hbs**

``` handlebars
<p>
    Me: "{{> @partial-block }}", <br />
    Echo: "{{> @partial-block }}"
</p>
```

**demo.hbs**

``` handlebars
<div>
    {{#> template}}
        Hello echo!
    {{/template}}
</div>
```

Currently no possibility to check, but haven't touched  the code and with with v4.0.3 it seemed to work, however after commit 7535e48 it produces the following console output:

``` bash
events.js:160
      throw er; // Unhandled 'error' event
      ^
Error: The partial @partial-block could not be found
    at Object.invokePartial (<PROJEKT_ROOT>/node_modules/handlebars/dist/cjs/handlebars/runtime.js:255:11)
    at Object.invokePartialWrapper [as invokePartial] (<PROJEKT_ROOT>/node_modules/handlebars/dist/cjs/handlebars/runtime.js:65:39)
    at Object.eval (eval at createFunctionContext (<PROJEKT_ROOT>/node_modules/handlebars/dist/cjs/handlebars/compiler/javascript-compiler.js:252:23), <anonymous>:14:28)
    at main (<PROJEKT_ROOT>/node_modules/handlebars/dist/cjs/handlebars/runtime.js:170:32)
    at Object.ret (<PROJEKT_ROOT>/node_modules/handlebars/dist/cjs/handlebars/runtime.js:173:12)
    at Object.lsg-demo-section (<PROJEKT_ROOT>/node_modules/handlebars/dist/cjs/handlebars/compiler/compiler.js:467:21)
    at Object.invokePartialWrapper [as invokePartial] (<PROJEKT_ROOT>/node_modules/handlebars/dist/cjs/handlebars/runtime.js:69:46)
    at eval (eval at createFunctionContext (<PROJEKT_ROOT>/node_modules/handlebars/dist/cjs/handlebars/compiler/javascript-compiler.js:252:23), <anonymous>:5:31)
    at prog (<PROJEKT_ROOT>/node_modules/handlebars/dist/cjs/handlebars/runtime.js:206:12)
    at execIteration (<PROJEKT_ROOT>/node_modules/handlebars/dist/cjs/handlebars/helpers/each.js:42:19)
```

/cc @lawnsea

nknapp:
[...]

Issue #1185, opened 2016-02-18
Stack blown when nested "@partial-block"-using partial not given a partial block.


damncabbage:
I just hit this in the wild with someone accidentally calling a `@partial-block`-using partial without giving it said partial block.

Minimal example:

```
var Handlebars = require('handlebars');
Handlebars.registerPartial('outer', '<p>{{> @partial-block }}</p>');
Handlebars.registerPartial('inner', '<span>{{> @partial-block }}</span>');

// Calls "inner" without giving it a partial block.
Handlebars.compile("<div>{{#>outer}} {{>inner}} {{/outer}}</div>")(); 
```

Produces:

```
RangeError: Maximum call stack size exceeded
    at Object.eval (eval at createFunctionContext (…/node_modules/handlebars/dist/cjs/handlebars/compiler/javascript-compiler.js:254:23), <anonymous>:1:10)
    at main (…/node_modules/handlebars/dist/cjs/handlebars/runtime.js:173:32)
    at ret (…/node_modules/handlebars/dist/cjs/handlebars/runtime.js:176:12)
    at ret (…/node_modules/handlebars/dist/cjs/handlebars/compiler/compiler.js:525:21)
    at Object.invokePartial (…/node_modules/handlebars/dist/cjs/handlebars/runtime.js:268:12)
    at Object.invokePartialWrapper [as invokePartial] (…/node_modules/handlebars/dist/cjs/handlebars/runtime.js:68:39)
    at eval (eval at createFunctionContext (…/node_modules/handlebars/dist/cjs/handlebars/compiler/javascript-compiler.js:254:23), <anonymous>:6:28)
    at prog (…/node_modules/handlebars/dist/cjs/handlebars/runtime.js:219:12)
    at Object.invokePartial (…/node_modules/handlebars/dist/cjs/handlebars/runtime.js:268:12)
    at Object.invokePartialWrapper [as invokePartial] (…/node_modules/handlebars/dist/cjs/handlebars/runtime.js:68:39)
```

As far as I can tell: the `inner` partial, when not giving a partial-block of its own, will refer to the `@partial-block` given to the `outer` partial.

Tested with Handlebars 4.0.5.

stringbean:
This also affects inline partials. [JSFiddle](https://jsfiddle.net/stringbean/9dkmhs5w/) showing this.

jaichandra:
Nested partials is critical to building modular templates. This needs to be addressed on priority.

[...]

PR #1290, opened 2017-01-01
WIP: Possible fix for #1252: Using @partial-block twice in a template not possible

nknapp:
I've added a possible fix for this bug, but it includes some refactorings of the `@partial-block`-code. After thinking about it a lot (this is kind of a brain-twister), I've come to the conclusions that partial-blocks should be treated like function-closures:

With the example from one of the tests:

### Template 

```hbs
<template>{{#> outer}}{{value}}{{/outer}}</template>
```

### outer

```hbs
<outer>{{#> nested}}<outer-block>{{> @partial-block}}</outer-block>{{/nested}}</outer>
```

### inner

```hbs
<nested>{{> @partial-block}}</nested>
```

### Explanation

When the partial-block of the outer partial  (`<outer-block>{{> @partial-block}}</outer-block>`) is executed, the data-variable `@partial-block` of `outer` must be called. The fix does not achieve this by popping the partial-block once its been used, but by storing the partial-block in the local closure-context of the `invokePartial`-function. The partial-block for that partial is than wrapped with another function that loads the previous partial-block from the local closure-context.

I can add more explanation, once my laptop-battery has recharged. In any way, I would like to have a review of the changes. 

@lawnsea I changed some of the code from your fix for #1185, so could you review if my changes are valid in your eyes? The changes are in the branch "tmp/issue-1252".

After review, I would probably make some style changes for better readability.

### DIFF
--- a/lib/handlebars/runtime.js
+++ b/lib/handlebars/runtime.js
@@ -210,12 +210,7 @@ export function wrapProgram(container, i, fn, data, declaredBlockParams, blockPa
 export function resolvePartial(partial, context, options) {
   if (!partial) {
     if (options.name === '@partial-block') {
-      let data = options.data;
-      while (data['partial-block'] === noop) {
-        data = data._parent;
-      }
-      partial = data['partial-block'];
-      data['partial-block'] = noop;
+      partial = options.data['partial-block'];
     } else {
       partial = options.partials[options.name];
     }
@@ -228,6 +223,8 @@ export function resolvePartial(partial, context, options) {
 }
 
 export function invokePartial(partial, context, options) {
+  // Use the current closure context to save the partial-block if this partial
+  const currentPartialBlock = options.data && options.data['partial-block'];
   options.partial = true;
   if (options.ids) {
     options.data.contextPath = options.ids[0] || options.data.contextPath;
@@ -236,10 +233,17 @@ export function invokePartial(partial, context, options) {
   let partialBlock;
   if (options.fn && options.fn !== noop) {
     options.data = createFrame(options.data);
-    partialBlock = options.data['partial-block'] = options.fn;
-
-    if (partialBlock.partials) {
-      options.partials = Utils.extend({}, options.partials, partialBlock.partials);
+    // Wrapper function to get access to currentPartialBlock from the closure
+    let fn = options.fn;
+    partialBlock = options.data['partial-block'] = function partialBlockWrapper(context, options) {
+      // Restore the partial-block from the closure for the execution of the block
+      // i.e. the part inside the block of the partial call.
+      options.data = createFrame(options.data);
+      options.data['partial-block'] = currentPartialBlock;
+      return fn(context, options);
+    };
+    if (fn.partials) {
+      options.partials = Utils.extend({}, options.partials, fn.partials);
     }
   }
 
--- a/spec/partials.js
+++ b/spec/partials.js
@@ -249,6 +249,13 @@ describe('partials', function() {
         true,
         'success');
     });
+    it('should be able to render the partial-block twice', function() {
+      shouldCompileToWithPartials(
+          '{{#> dude}}success{{/dude}}',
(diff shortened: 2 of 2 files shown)


lib/handlebars/runtime.js
```diff
@@ -236,10 +233,17 @@ export function invokePartial(partial, context, options) {
...
+    let fn = options.fn;
+    partialBlock = options.data['partial-block'] = function partialBlockWrapper(context, options) {
+      // Restore the partial-block from the closure for the execution of the block
+      // i.e. the part inside the block of the partial call.
+      options.data = createFrame(options.data);
+      options.data['partial-block'] = currentPartialBlock;
+      return fn(context, options);
+    };
+    if (fn.partials) {
+      options.partials = Utils.extend({}, options.partials, fn.partials);
```
lawnsea: So, previously, `fn === partialBlock`. Are you sure that we don't depend on that elsewhere?

lawnsea:
I am not sure the language should allow multiple `@partial-block`s at a given nesting level. I'd like some feedback from the other maintainers and @wycats in particular.

spec/partials.js
```diff
@@ -249,6 +249,13 @@ describe('partials', function() {
         true,
         'success');
     });
+    it('should be able to render the partial-block twice', function() {
+      shouldCompileToWithPartials(
+          '{{#> dude}}success{{/dude}}',
+          [{}, {}, {dude: '{{> @partial-block }} {{> @partial-block }}'}],
```
nknapp: I can add those tests.

lib/handlebars/runtime.js
nknapp: I have searched the `lib`-folder and  the `src`-folder for occurences of the string `partial-block` in the `tmp/issue-1252`-branch and the only matches are in the functions `invokePartial` and `resolvePartial`. Those functions do not check for `fn === partialBlock`.

But no, I cannot be completely sure. I have just started digging in the internals of Handlebars, there might be something I don't see. I'm pretty confident though.

nknapp:
You also need this patch, if you want to use the `#each`-helper with partial-blocks

* Template: `<template>{{#> list value}}value = {{.}}{{/list}}</template>`
* Partial "list": `<list>{{#each .}}<item>{{> @partial-block}}</item>{{/each}}</list>`
* Input: `{ value: [ 'a', 'b', 'c' ] }`

nknapp:
@wycats The question we are having here is:

Should this be possible: `<partial>{{@partial-block}} - {{@partial-block}}</partial>`
and this: `<partial>{{#each list}}{{@partial-block}}{{/each}}</partial>`

It was possible in version 4.0.3 and 4.0.4, but there were other problems (#1185). Versions 4.0.5 and 4.0.6 do not allow it.

You said you don't want new features to be implemented until we have a spec. Do you consider the above examples to be a new feature or a fix? 
My opinion is that it is a fix. Since a `@partial-block` is basically the equivalent of the `options.fn()`-callback of a block-helper, it should work like `options.fn()`, even if it is called multiple times.

@lawnsea I have added multiple test-cases for your scenarios and a few more (e.g. a partial-block called from within an `#each` block. Those tests pass.

lawnsea:
@nknapp I've come around on this, particularly since it used to work and doesn't anymore, thanks to me. I still think `<partial>{{@partial-block}} - {{@partial-block}}</partial>` is weird and not very compelling, but `<partial>{{#each list}}{{@partial-block}}{{/each}}</partial>` is pretty clearly useful.

Thanks for adding the test cases. I'll take a look.

lawnsea:
Let's give @wycats a couple more days to weigh in. If we don't hear from him, I think we should land this and cut a 4.0.7 release.

nknapp:
This has been waiting for almost a month now. I have not had the time to do it, but I will merge it now.

nknapp:
I have condensed the branch into 2 commits
- The fix with a test for the fix
- Additional testcases for partial-blocks

Issue #141, opened 2011-11-18
Are async handlers on the roadmap?


ghost:
I'd love to be able to register handlers that run async functions and have the results of those functions be rendered into the template. There aren't very many options for that right now, but dust and asyncEJS both have interesting mechanisms that support it. Is this anywhere on the radar or is it just totally out of scope for Handlebars?

felix

PR #1293, opened 2017-01-05
Support async #with contexts

andreas-gruenbacher:
In ```{{#with context}}body{{/with}}```, when context is a function, the
functions's return value is used as the context in body.  This allows
for computed context, but not when that context is provided
asynchronously, such as when querying a database.

To support asynchronous contexts as well, check if context is a function
that takes arguments.  If it is, call it with the callback to invoke
when done as the first, and ```#with```'s hash arguments as the second
argument.  For example, with the following context function:
```
 person: function(fn) {
    fn({first: 'Alan', last: 'Johnson'});
  }
```
The template:
```
  {{#with person}}{{first}} {{last}}{{/with}}
```
will render to:
```
  Alan Johnson
```

### DIFF
--- a/lib/handlebars/helpers/with.js
+++ b/lib/handlebars/helpers/with.js
@@ -2,19 +2,29 @@ import {isEmpty, isFunction} from '../utils';
 
 export default function(instance) {
   instance.registerHelper('with', function(context, options) {
-    if (isFunction(context)) { context = context.call(this); }
-
-    let fn = options.fn;
-
-    if (!isEmpty(context)) {
-      let data = options.data;
-
-      return fn(context, {
-        data: data,
-        blockParams: [context]
-      });
+    let async;
+    if (isFunction(context) && context.length) {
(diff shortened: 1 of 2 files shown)


andreas-gruenbacher:
A similar approach would work for #each as well, of course.

andreas-gruenbacher:
This is unfortunately implemented wrong so far; need to figure out how to fix it.

andreas-gruenbacher:
I see that asynchronous helpers have been discussed before (https://github.com/wycats/handlebars.js/issues/141), and that people were not very interested. So sorry for the noise.

andreas-gruenbacher:
@nknapp thanks. I've had a quick look at 'promised-handlebars' and the placeholder approach seems so awfully wrong that I'd much rather switch to a different template engine that supports async helpers in the long run. For now, I think I'll just resort to doing all possible database queries ahead of the rendering.

PR #1295, opened 2017-01-13
Moved .eslintrc to .eslintrc.js and cleaned up rules

travnels:
This PR addresses a couple items around ESLint configurations:
- According to http://eslint.org/docs/user-guide/configuring#configuration-file-formats, the `.eslintrc` file is deprecated. There are several file types that can replace `.eslintrc`, with `.eslintrc.js` being the top priority. 
- To reduce some of the clutter in this file, I extended from `eslint:recommended`, which enables a subset of core rules that report common problems. A list of these rules can be found on the eslint rules page, http://eslint.org/docs/rules/. 
- For better readability with the eslint config file, I renamed the rules settings per http://eslint.org/docs/user-guide/configuring#configuring-rules:
   `0` -> `off`
   `1` -> `warn`
   `2` -> `error`

Only a change to the eslint configuration
Focused on getting a eslint configuration that is identical to the current rules, but will keep up to date by extending the `eslint:recommended`
- [x] Have tests 
Verified the `grunt eslint` ran clean after all the changes were made to the eslint config file.


nknapp:
This must have been some tidious work. Those rules you omitted are all part of the "recommended" ruleset? I just want to be sure that no rules are turned off now, that where previously turned on.

Replacing "2" with "error" adds a lot of clarity.

travnels:
Yes, the omitted rules are part of `"extends": "eslint:recommended"` and can be found on http://eslint.org/docs/rules/ (the rules with a check mark). I reviewed each rule one by one to make sure it could be removed. In doing so, it identified a list of rules that are overrides from the "recommended" set (indicated as `// overrides eslint:recommended defaults` in the file). These rules might want to be addressed at a later time, but I didn't want to include those as part of this PR, since that deviates from the `focused on a single change` you like to see.

Issue #1227, opened 2016-05-26
#each is not working for `new Array()` where as it is working for `[ ]`


Swathi8:
I have case where I need to be looping through empty array of length 2.

If I pass empty array as `[undefined,undefine]` or `[null, null]`, each works fine but if i pass array as `new Array(2)` looping is not working.

jsfiddle: https://jsfiddle.net/SwathiP/v3hsc2qf/1/

lawnsea:
This one might be a bug. Labelling it as such for now. Feel free to overrule me, fellow collaborators.

nknapp:
#1065

travnels:
This should be addressed now in https://github.com/wycats/handlebars.js/pull/1299

nknapp:
I am not sure what the correct behavior should be here. I had not seen #1093 before:

* #1065 was the initial issue: The jsfiddle at http://jsfiddle.net/InoKu/5ybzdbye/1/ demonstrates how with version 1.0.rc.2 and without "use strict", `this.Number` resolves the the global "Number()"-function when called from a block-helper on an undefined context. The solution to this was 06d515a89d18b50805a5fe4eec8f1156bbe92d45 on Aug 4, 2015
* In #1093 the behavior was partially reverted (fffb5a985f8ff7cd75f2c16b37b69f7ddd3ba9a0) such that missing entries are not iterated, but `undefined` entries are not. (Sep 24, 2015).

http://jsfiddle.net/5ybzdbye/3/ shows that the original problem of #1065 is not present in Handlebars 4.0.6. 

I guess I have to revert my previous statement. It seems to me like this behavior is wanted:

* Sparse arrays and deleted entries (like `var a = [1,2,3]; delete a[2];`) are ignored during iteration.
* Values that are set to undefined explicitly (like `var=[1,2,3]; a[1]=undefined`) are iterated.

Until [we have spec](https://github.com/wycats/handlebars.js/issues/1277), I would not change this. 

 (I'm relatively new here as a collaborator, so forgive that I didn't come up with this immediately.

nknapp:
This is a copy of my last comment on the PR, which explains why I set this to "docs-needed" rather than "bug".

The current behaviour of Handlebars is this (https://jsfiddle.net/9D88g/70/):

When iterating an object, keys with undefined values will be iterated (due to #1093)
When iterating an array where entries have been deleted with delete cols[1] (due to #1065).
[...]

Issue #1065, opened 2015-07-24
Each iteration over sparse arrays incorrectly executes on deleted entries


BackupForce:
I got an arraydata like this { rocords:[{Number:'1',Name:'test1'},{Number:'2',Name:'test2},{Number:'3',Name:'test3}] };
registerHelper like this
Handlebars.registerHelper('link', function (options) {
        return '<p>' + options.fn(this) + this.Name +'</p>';
});
I delete array[1];
when registerHelper execute array[1], 'this' ponit window.
Its something wrong?

kpdecker:
I'm not clear what your issue is here and a live example would help us debug this in a much better fashion. Please create a jsfiddle showing us the unexpected behavior. The contributing doc links to a template that can be used for this.

BackupForce:
finally haha
http://jsfiddle.net/InoKu/5ybzdbye/1/

kpdecker:
Thanks for the fiddle, it looks like we are iterating over sparse arrays incorrectly. We need to check to see the field exists prior to running it as an undefined context is not supported (outside of javascript strict mode). Will be fixed in the next release.

Issue #1093, opened 2015-09-08
#each skips entry of dict with empty value since 4.0.0


florianpilz:
If I have an template like this:

    {{#each data}}
    <span class="{{@key}}">{{this}}</span>
    {{/each}}

Handlebars will render nothing when data is a dict entry with empty value, i.e. if I render above template with

    template({data: {
        'first': null,
        'second': 'bar',
    }});

Only `second` will be rendered, but `first` will be ignored.

I have made a JSFiddle to demonstrate the behaviour:

* old behaviour with Handlebars 3.0.3, where it will **render** `first` https://jsfiddle.net/jag017t2/1/
* new behaviour with Handlebars 4.0.2, where it will **skip** `first` https://jsfiddle.net/5n6t61kz/5/

I have tested the behaviour with 4.0.0 and it is skipped there as well, so it's a problem introduced in 4.0.0.

Is this the intended behaviour now? It is not stated in the documentation and I can't see a good match in the release notes. It breaks our application, since we expect to have 7 `<td>` columns rendered by Handlebars, but 2 are missing due to `null` values.

kpdecker:
This was changed due to #1065 where helpers fail when executing against a null context for non-strict mode. I would suggest using `''` rather than `null` as this will render safely for all contexts. Baring that we might be able to implement a flag to allow for this unsafe iteration but I'm not sure on what sort of ETA that might have.

Marsup:
Same problem with undefined values on lout for me where I try to iterate over objects.
I'm not sure I see why it's related to sparse arrays, can't we make a difference between objects and arrays ?

kpdecker:
@Marsup the root problem is that helpers were being called like `helper.call(undefined)` and under non-strict mode this would execute against the global object leading to surprises. I'm now starting to think that it's better to:

1. Execute for all defined keys (sparse arrays will skip elements)
2. Fail over context execution to an empty object if not defined. i.e. something like `helper.call(context || {})`

It's not ideal but seems to provide the best middle ground for handling undefined in non-strict mode.

Marsup:
You should probably explicitly check for null and undefined, other falsy values should work.

kpdecker:
Good call, took that into account.

kpdecker:
Released in 4.0.3

florianpilz:
Thanks, works like a charm!

PR #1299, opened 2017-01-16
Handle undefined arrays better

travnels:
Addressing issue identified in https://github.com/wycats/handlebars.js/issues/1227 with arrays that have undefined elements.

[undefined, undefined] should have the same behavior as new Array(2),
per https://github.com/wycats/handlebars.js/issues/1065

- [x] Have tests

### DIFF
--- a/lib/handlebars/helpers/each.js
+++ b/lib/handlebars/helpers/each.js
@@ -36,7 +36,7 @@ export default function(instance) {
     if (context && typeof context === 'object') {
       if (isArray(context)) {
         for (let j = context.length; i < j; i++) {
-          if (i in context) {
+          if (context[i] !== undefined) {
             execIteration(i, i, i === context.length - 1);
           }
         }
--- a/spec/regressions.js
+++ b/spec/regressions.js
@@ -288,4 +288,11 @@ describe('Regressions', function() {
 
     shouldCompileTo('{{helpa length="foo"}}', [obj, helpers], 'foo');
   });
+
+  it('GH-1227: Array of undefineds', function() {
+    var array = [undefined, undefined];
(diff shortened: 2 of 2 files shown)


nknapp:
As I said in #1227: Despite what I said earlier, this would probably re-introduce #1093 and is probably not the desired behavior.

travnels:
The regression tests for #1093 still pass with this PR (`GH-1093: Undefined helper context`, https://github.com/travnels/handlebars.js/blob/array-of-undefineds/spec/regressions.js#L207-L224). Is there more to the regressions tests we should add?

nknapp:
Ah, #1093 was about objects, not arrays. That's why the regression test still passes.
Still, it's not easy to tell right from wrong here. See this jsfiddle: https://jsfiddle.net/9D88g/65/

* When iterating an object, keys with `undefined` values will be iterated (due to #1093)
* When iterating an array where entries have been deleted with `delete cols[1]` (due to #1065).

It is not explicitly defined how `['a',undefined,'b']` should behave (that's why we need a spec), but the current behavior is consistent between objects and arrays. 
And it is consistent with the `Array.prototype.forEach`:

```js
['a',undefined,'c'].forEach((i) => console.log('a',i));
// => a a
// => a undefined
// => a c
var x= ['a','b','c'];
delete x[1];
x.forEach((i) => console.log('a',i));
// => a a
// => a c
``` 
There is no critical need for this change: Anyone can implement a custom block-helper that iterates the values differently. And I think, it is not unlikely that changing this would break somebodies code.

travnels:
I agree, a spec for this is necessary. I also agree that the consistency with `Array.prototype.forEach` is probably correct. Perhaps we should change the label for #1227 from `bug` to `docs-needed`?

nknapp:
I'm closing this PR for now.

PR #1308, opened 2017-02-13
Improve source map support

joonas-lahtinen:
Fix duplicated sourceMappingURL line when generating a map AND minifying. Add support for inlining source maps (wasn't quite sure if you'll prefer --inline-map, --map-inline, --embed-map or what). Improve the tests to reflect new features.


joonas-lahtinen:
Hmm, bit confused as locally it's all passing. Trying one by one.

PR #1310, opened 2017-02-14
Avoid duplicate "sourceMappingURL=" lines.

joonas-lahtinen:
Avoid duplicate // sourceMappingURL=... lines when minifying AND
generating a map. UglifyJS2 will write the line when minifying.


nknapp:
This should also be merged to 4.x

PR #1315, opened 2017-02-23
parser: Change suffix to use ES6 default module export

### LINKED ISSUE #1313: Distribute original ES6 sources
As suggested in https://github.com/emberjs/ember.js/pull/14853#discussion_r101679795 it would be awesome if the original ES6 module sources would be included in the distributed package.

Turbo87:
This export will be transpiled by Babel for the `cjs` distribution, but will enable others to use a pure ES6 module distribution.

Resolves #1313 

/cc @chadhietala @mmun

### DIFF
--- a/.istanbul.yml
+++ b/.istanbul.yml
@@ -1,2 +1,2 @@
 instrumentation:
-    excludes: ['**/spec/**']
+    excludes: ['**/spec/**', '**/handlebars/compiler/parser.js']
--- a/src/parser-prefix.js
+++ b/src/parser-prefix.js
@@ -1,1 +1,1 @@
-/* istanbul ignore next */
+// File ignored in coverage tests via setting in .istanbul.yml
--- a/src/parser-suffix.js
+++ b/src/parser-suffix.js
@@ -1,2 +1,1 @@
-exports.__esModule = true;
-exports['default'] = handlebars;
+export default handlebars;


chadhietala:
This seems good. I wouldn't think the CI failures are related.

Turbo87:
yeah, that test failure seems to be unrelated

nknapp:
It's not completely unrelated. The reason is this:
```
ERROR: Coverage for statements (95.06%) does not meet global threshold (100%)
ERROR: Coverage for branches (95.77%) does not meet global threshold (100%)
ERROR: Coverage for functions (97.01%) does not meet global threshold (100%)
```
The generated file `dist/cjs/handlebars/compiler/parser.js` is not completely covered by tests.
There is a line `/* istanbul ignore next */` in the header but somehow it has no effect, when this PR is applied.

nknapp:
Coverage-Report of the parser.js file...

[parser-coverage.zip](https://github.com/wycats/handlebars.js/files/797823/parser-coverage.zip)

nknapp:
Since all the lines in this file have been ignored in the recent builds and I don't think it is possible to cover the entire file, I think it would be OK to exclude the file in [.istanbul.yml](https://github.com/wycats/handlebars.js/blob/master/.istanbul.yml).

Turbo87:
@nknapp good point. I've added the `parser.js` file to the istanbul ignore list

nknapp:
I have squashed the commits (makes it easier to cherry-pick to the 4.x-branch) and removed the now obsolete istanbul annotation from parser-prefix.

I hope you didn't mind me force-pushing to your es6-parser branch.

Turbo87:
Not at all, thanks!

PR #1321, opened 2017-03-09
Fix context-stack when calling block-helpers on null values

### LINKED ISSUE #1319: "unless" breaks when "each" value equals "null"
When using "each" over either an object or an array and a null value is encountered, "unless" will destroy the context, replacing the null value with an empty object, and dropping parent access all together.

Here is a reproduction test case: https://jsfiddle.net/r33cjjb2/3/

To aid seeing the issue, the failing case renders the following inline: !!!FAIL!!!FAIL!!!FAIL!!!FAIL!!!FAIL!!!FAIL!!!FAIL!!!FAIL!!!FAIL!!!FAIL!!

Looking forward to seeing this fixed.  It is a real pain in my @$$ atm.  Otherwise thank you for a magnificent tool.

nknapp:
Fixes #1319

Original behaviour:
- When a block-helper was called on a null-context, an empty object was used
  as context instead. (#1093)
- The runtime verifies that whether the current context equals the
  last context and adds the current context to the stack, if it is not.
  This is done, so that inside a block-helper, the ".." path can be used
  to go back to the parent element.
- If the helper is called on a "null" element, the context was added, even
  though it shouldn't be, because the "null != {}"

Fix:
- The commit replaces "null" by the identifiable "container.nullContext"
  instead of "{}". "nullContext" is a sealed empty object.
- An additional check in the runtime verifies that the context is
  only added to the stack, if it is not the nullContext.

### DIFF
--- a/lib/handlebars/compiler/javascript-compiler.js
+++ b/lib/handlebars/compiler/javascript-compiler.js
@@ -987,7 +987,7 @@ JavaScriptCompiler.prototype = {
     let params = [],
         paramsInit = this.setupHelperArgs(name, paramSize, params, blockHelper);
     let foundHelper = this.nameLookup('helpers', name, 'helper'),
-        callContext = this.aliasable(`${this.contextName(0)} != null ? ${this.contextName(0)} : {}`);
+        callContext = this.aliasable(`${this.contextName(0)} != null ? ${this.contextName(0)} : (container.nullContext || {})`);
 
     return {
       params: params,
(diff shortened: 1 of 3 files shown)


nknapp:
I'd like to have a short review by any other collaborator.

nknapp:
I am not sure, but the compiled template now uses a property of the container (`nullContext`) that needs to be present in the runtime.
Do I have to increase in the compiler-version? Is this considered to be a breaking change?

nknapp:
I have changed the behavior such that compiled templates >=4.0.7 are compatible with runtimes <4.0.7, so it's not breaking change anymore. I will now merge this into 4.x and master

Issue #1324, opened 2017-03-13
Undocumented syntactic difference with mustache


robinhouston:
I’m in the process of converting a project from mustache to handlebars, and I ran into a syntactic difference that isn’t documented in [the differences section](https://github.com/wycats/handlebars.js#differences-between-handlebarsjs-and-mustache).

Mustache allows whitespace between the `{{` and the `#`, `>`, `^`, etc., which Handlebars does not. For example this snippet:
```{{ # user }}{{ > header_logged_in }}{{ / user }}
{{ ^ user }}{{ > header_not_logged_in }}{{ / user }}
```
had to be changed to:
```{{# user }}{{> header_logged_in }}{{/ user }}
{{^ user }}{{> header_not_logged_in }}{{/ user }}
```
to work with Handlebars.

It may be useful to mention this in the docs.

nknapp:
Hi. It would be great if you could create a PR that adds a bullet-point to [the appropriate section](https://github.com/wycats/handlebars.js#compatibility). This certainly is noteworthy.

PR #1325, opened 2017-03-13
Document whitespace compatibility difference with mustache

robinhouston:
See wycats/handlebars.js#1324


nknapp:
Awesome, thanks.

PR #1330, opened 2017-04-09
Added -w, --watch option to the handlebars CLI

m-ahmadi:
Watch option for the handlebars CLI.
It watches for changes on the input file(s) / dir(s), and performs reprecompilation on changes to them.
I used the 'chokidar' package for watching files and folders, and the 'color' package for outputing nice looking messages to the console.

### DIFF
--- a/bin/handlebars
+++ b/bin/handlebars
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-var optimist = require('optimist')
+var yargs = require('yargs')
     .usage('Precompile handlebar templates.\nUsage: $0 [template|directory]...', {
       'f': {
         'type': 'string',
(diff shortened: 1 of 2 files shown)


m-ahmadi:
If you don't want the ES6 stuff, tell me to replace them.

nknapp:
I personally feel a bit controversial about this request. "chokidar" adds another 2MB to the total size of Handlebars. I have a few projects that use Handlebars and I am constantly trying to *reduce* the size of those projects. 
I think that such a development tool should be in a separate project.

stevenvachon:
This should probably be handled with [npm-watch](https://npmjs.com/npm-watch), since it can restart the process, which is particularly useful if the current process is not yet complete.

m-ahmadi:
I made another pr and changed it from two commits into one, can you guys please copy-paste your comments there?

PR #1332, opened 2017-04-09
Added watch option (-w, --watch) for the handlebars cli.

m-ahmadi:
It watches the input file(s) / dir(s), and on changes to them, it performs the precompilation all over again.
I used the 'chokidar' package for watching files and folders, and the 'color' package for outputing nice looking messages to the console.
If you don't want the ES6 stuff tell me to replace them.

### DIFF
--- a/bin/handlebars
+++ b/bin/handlebars
@@ -95,6 +95,11 @@ var yargs = require('yargs')
         'description': 'Prints the current compiler version',
         'alias': 'version'
       },
+      'w': {
+        'type': 'boolean',
+        'description': 'Watch the input file(s)/dir(s) and perform precompilation on changes.',
+        'alias': 'watch'
+      },
 
       'help': {
         'type': 'boolean',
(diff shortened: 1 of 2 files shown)


nknapp:
I personally feel a bit controversial about this request. "chokidar" adds another 2MB to the total size of Handlebars. I have a few projects that use Handlebars and I am constantly trying to reduce the size of those projects.
I think that such a development tool should be in a separate project.

But, this is my opinion only. There may be different ones.

m-ahmadi:
@nknapp Actually that bothered me too, because I was thinking the same thing, that they probably are not going to like this because there's two additional packages, So I thought it, but then damn using 'chikidar' is so easy, and I didn't have that much time, So I just wanted to put something together that works, I do this task (watching templates), a lot so I thought it would be nice if it was in the package itself.

nknapp:
I'm closing this. If anyone really really needs this feature, please add a comment.

PR #1349, opened 2017-05-18
Handlebars.compile() does not modify "options" anymore

### LINKED ISSUE #1327: each overwrites data in the "data" property.
I am using `handlebars` as the templating engine of a `gulp-metalsmith` build system.  I am  using the `metalsmith-data-loader` to load a JSON file into the a `data` property; however, when I go to access that property in then handlebars template it is not the value of `true`.  It seems to be linked to the access of the `@index` in an each block helper.

I presume the problem is in: `handlebars.js/lib/handlebars/helpers/each.js` however, I don't know enough about the expected params to really grok the problem.

nknapp:
Fixes #1327

- This commit creates a shallow copy of the "options" passed to
  Handlebars.compile() in order to prevent modifications
- Note that "new Handlebars.Compiler().compile(..., options)" still
  modify the options object. This might change in the future, if
  anybody needs a fix for that.

### DIFF
--- a/lib/handlebars/compiler/compiler.js
+++ b/lib/handlebars/compiler/compiler.js
@@ -70,7 +70,7 @@ Compiler.prototype = {
       for (let name in knownHelpers) {
         /* istanbul ignore else */
         if (name in knownHelpers) {
-          options.knownHelpers[name] = knownHelpers[name];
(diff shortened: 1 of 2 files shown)


nknapp:
AppveyorCI is not configured yet, thus the failure

PR #1370, opened 2017-07-31
Fix Commits links in release-notes

radium-v:
The compare links in `release-notes.md` were linking to a different repo. They should be pointing to base repo to work correctly.


nknapp:
Thanks, that's probably because I have the "origin"-remote pointed to my fork. I'll have to check before the next release.

PR #1392, opened 2017-10-11
move uglify-js to dependencies

### LINKED ISSUE #1391: `uglify-js` is unconditionally imported, but only listed as optional dependency
This is causing https://github.com/ember-cli/ember-cli/issues/7371, where people are using `npm` to install `handlebars` as a transitive dependency but apparently `npm` is not installing the optional `uglify-js` dependency (anymore?) which makes the import fail.

kellyselden:
Closes #1391

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,9 +23,7 @@
   "dependencies": {
     "async": "^1.4.0",
     "optimist": "^0.6.1",
-    "source-map": "^0.4.4"
-  },
-  "optionalDependencies": {
+    "source-map": "^0.4.4",
     "uglify-js": "^2.6"
   },
   "devDependencies": {


nknapp:
Personally, I'm in favour of keeping the dependency optional and handle the missing library case. It greatly reduces the amount of dependencies and downloaded megabytes if you don't need the precompiler. Opinions?

rwjblue:
Agreed, but with the move to ES modules how do we conditionally import?

nknapp:
I would make an execption here. It may make sense to move to completely towards es6-modules, but if they lack support for essential features, we should omit them when we can't use them.

I would suggest using `require('uglify-js')` instead and adding a lengthy comment, with the reason for not using es6.

nknapp:
I made a PR here: #1394

nknapp:
I'm closing this in favor of #1394

PR #1393, opened 2017-10-11
Add jsDelivr hits badge

LukasDrgon:
Hey, I noticed that handlebars.js gets 19M hits per month on jsDelivr, so I thought you might like this badge.


nknapp:
There seem to be no objections...

PR #1394, opened 2017-10-13
Gracefully handle missing uglify-js dependency

### LINKED ISSUE #1391: `uglify-js` is unconditionally imported, but only listed as optional dependency

nknapp:
closes #1391

uglify-js is an optional dependency and should be treated as such.
This commit gracefully handles MODULE_NOT_FOUND errors while loading
uglify.

### DIFF
--- a/lib/precompiler.js
+++ b/lib/precompiler.js
@@ -4,7 +4,7 @@ import fs from 'fs';
 import * as Handlebars from './handlebars';
 import {basename} from 'path';
 import {SourceMapConsumer, SourceNode} from 'source-map';
-import uglify from 'uglify-js';
+
 
 module.exports.loadTemplates = function(opts, callback) {
   loadStrings(opts, function(err, strings) {
@@ -235,7 +235,6 @@ module.exports.cli = function(opts) {
     }
   }
 
-
   if (opts.map) {
     output.add('\n//# sourceMappingURL=' + opts.map + '\n');
   }
@@ -244,12 +243,7 @@ module.exports.cli = function(opts) {
   output.map = output.map + '';
 
   if (opts.min) {
-    output = uglify.minify(output.code, {
-      fromString: true,
-
-      outSourceMap: opts.map,
-      inSourceMap: JSON.parse(output.map)
-    });
+    output = minify(output, opts.map);
   }
 
   if (opts.map) {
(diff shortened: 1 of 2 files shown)


nknapp:
I'll merge this request to 4.x and master on Monday evening (Europe) if nobody objects. Please let me know if you need it earlier.

lib/precompiler.js
```diff
@@ -4,7 +4,17 @@ import fs from 'fs';
...
+// We are using `require` instead of `import` here, because es6-modules do not allow
+// dynamic imports and uglify-js is an optional dependency.
+var uglify;
+try {
+  uglify = require('uglify-js');
+} catch (err) {
+  if (err.code !== 'MODULE_NOT_FOUND') {
+    throw err;
+  }
+}
```
Turbo87: @nknapp it might be easier to move the `require()` call into the `if (opts.min) {` condition block below. that way you will only try to import the dependency if someone is actually trying to use the `min` option and you won't need the try-catch at all.

Turbo87:
@nknapp thanks a lot for looking into this. we've addressed this issue on the Ember CLI end already too so this is not time critical anymore, but probably still good to fix it :)

nknapp:
I thought it would be better for the readability of the code if the require was at the beginning. But you are probably right. There is no need to load uglify when the user does not want to minify in the first place.

nknapp:
Have added another commit and I will squash them before merging. Thanks for reviewing :smirk:

lib/precompiler.js
```diff
@@ -244,12 +258,19 @@ module.exports.cli = function(opts) {
...
+      // dynamic imports and uglify-js is an optional dependency.
+      let uglify = require('uglify-js');
+      output = uglify.minify(output.code, {
+        fromString: true,
+        outSourceMap: opts.map,
+        inSourceMap: JSON.parse(output.map)
+      });
+    } else {
+      console.error('Code minimization is disabled due to missing uglify-js dependency');
+    }
```
Turbo87: why not just use try/catch here directly instead of if/else and the function above?

nknapp: I want to use require.resolve to detect if uglify exists, beause it wont be irritated by other missing-module-errors (caused by dependencies of uglify).

I could put the `.minify`-call into the try-catch-block, but that would be something like 

A function seems to be the best way to create a try-catch construct 

```js
let exists;
try {
  require.resolve('uglify-js');
  exists = true
} catch (e) {
  if (e.code !== 'MODULE_NOT_FOUND') {
    throw e;
  }
  exists = false
}
if (exists) {
  require('uglify-js').minify(...)
}
```

This seems to be more complicated and I don't particular like the use of the `exists` variable in this code. Creating a function seems to be more readable.
But I think I will still make a change.

PR #1396, opened 2017-10-17
Update jsfiddle-link to a template using 4.0.11

nknapp:
Update the jsfiddle link after publish of 4.0.11. The new jsfiddle does not work yet, because the cdn has not caught 4.0.11 yet.


nknapp:
Push directly to master

Issue #1279, opened 2016-11-26
support bundling with rollup


pravi:
I'm trying to bundle handlebars with rollup (as webpack has a long list of dependencies), I get the following error. Can you support bundling with rollup?

```
$ rollup ./lib/handlebars.js 
'default' is not exported by lib/handlebars/compiler/parser.js (imported by lib/handlebars/compiler/base.js). For help fixing this error see https://github.com/rollup/rollup/wiki/Troubleshooting#name-is-not-exported-by-module
Error: 'default' is not exported by lib/handlebars/compiler/parser.js (imported by lib/handlebars/compiler/base.js). For help fixing this error see https://github.com/rollup/rollup/wiki/Troubleshooting#name-is-not-exported-by-module
    at Module.trace (/usr/lib/nodejs/rollup/src/Module.js:362:30)
    at Module.traceExport (/usr/lib/nodejs/rollup/src/Module.js:388:29)
    at Module.trace (/usr/lib/nodejs/rollup/src/Module.js:360:36)
    at ModuleScope.findDeclaration (/usr/lib/nodejs/rollup/src/ast/scopes/ModuleScope.js:47:22)
    at Scope.findDeclaration (/usr/lib/nodejs/rollup/src/ast/scopes/Scope.js:92:39)
    at Identifier.bind (/usr/lib/nodejs/rollup/src/ast/nodes/Identifier.js:7:29)
    at /usr/lib/nodejs/rollup/src/ast/Node.js:6:34
    at AssignmentExpression.eachChild (/usr/lib/nodejs/rollup/src/ast/Node.js:21:5)
    at AssignmentExpression.bind (/usr/lib/nodejs/rollup/src/ast/Node.js:6:8)
    at AssignmentExpression.bind (/usr/lib/nodejs/rollup/src/ast/nodes/AssignmentExpression.js:29:13)
Type rollup --help for help, or visit https://github.com/rollup/rollup/wiki

```

msroberts:
This is also an issue for me: because the compiled templates js file does not export anything, rollup does not modify the global Handlebars variable.

Related: https://github.com/wycats/handlebars.js/issues/1254, https://github.com/wycats/handlebars.js/issues/955

nknapp:
Could you try manually changing the last lines of `lib/handlebars/compiler/parser.js` from 

```
exports.__esModule = true;
exports['default'] = handlebars;
```
to
```
export default handlebars;
```

[...]

PR #1402, opened 2017-10-22
upgrade webpack

### LINKED ISSUE #1399: Update webpack to current version
We are still using webpack 1.x for development on Handlebars. The update means that changes to the configuration files must be done. (Tests are failing after just bumping the version)

timdeschryver:
Closes #1399


nknapp:
Hi and thanks for the PR. I pushed 2 commits to your branch. It seems that the babel packages need to be upgraded to version 6 and 7 in order to work with webpack 3, so I rebased my babel6-branch on your PR-branch.

tdeschryver:
Sorry, was a bit busy so I couldn't fix the build on lower versions of node.
It seems to me that the options `loose` and `optional` were both removed in babel 6 - [changelog](https://github.com/babel/babel/blob/master/CHANGELOG.md#600)

tdeschryver:
It is strange that some tests are failing...
I do think that `import Handlebars from './handlebars'` would be preferable? Because the default export is `Handlebars`.

I'll try to take a look this week.

tdeschryver:
Hmmm, I can't seem to fix it without breaking some tests.
Babel changed some things with the default export in babel 6? ([see here](https://blog.kentcdodds.com/misunderstanding-es6-modules-upgrading-babel-tears-and-a-solution-ad2d5ab93ce0)).

nknapp:
I see. There are two ways to get around this.

1. Identify all places where an "import * as foo from './bar.js'" is done inside the "lib/handlebars"-code and replace them with "import foo from './bar.js'". This also means changing the test-code in (spec/node.js) and it also means the the interface will be broken (because now you have to do `const Handlebars = require('handlebars').default` in node, rather than `require('Handlebars')`.
2. Identify all places where "export default" is done, and check if it actually needs to be `export { create, registerHelper, parse ... }.

We don't want to break the interface, so option to would be preferable, but 'm not sure that this really works and if this would not break anything too.

nknapp:
Those things I write here, currently feel like I'm given tips to you (@tdeschryver) with the expectation that you solve that problem. That is not my desire. It's great if you continue working on this. But since this is probably getting bigger than initially expected, I would like to give more people the opportunity to work on this and try things out.

I have pushed a branch `4.x-babel6` to allow others to place PRs against this branch with possible solutions.

tdeschryver:
Oh sorry, didn't see your messages.
I think I solved it but in a hacky way (because I'm not 100% sure how to export it properly)... :thinking:

nknapp:
The missing coverage seems to be in functions generated by babel (That's why travis is failing)

selection_001

The solution is tempting, but I don't know if tools that use ES6-import natively (like [rollup](https://rollupjs.org/)) will work with this fix (see #1279). This is not covered by tests, yet.

nknapp:
Mabye a combination will work. Using "export default" *and* "module.exports ="

tdeschryver:
The combination doesn't fix the code coverage. 
If this is still open I'll take a another look some days later.

tdeschryver:
OK if I let this one open for someone else? It seems like I'm overlooking something.
On a side note, I think the upgrade of webpack alone should be OK? Build errors seems to be related to `npm install`? If I remember correctly, I could build it locally on npm 4.

nknapp:
Sure, leave it open.

nknapp:
I've added the help-wanted badge. Maybe this will draw attention by others...

nknapp:
@tdeschryver I have rebased your branch onto the current 4.x and force-pushed it. I hope that does not mess up your history.

nknapp:
closing for now. We will have other attempts some time.

PR #1407, opened 2017-11-21
Ignore editor's files and build settings

aaharu:
handlebars@4.0.11 includes unnecessary files.

- [ ] Have tests

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,7 @@
 .DS_Store
 .gitignore
+.idea/*
+.istanbul.yml
 .rvmrc
 .eslintrc
 .travis.yml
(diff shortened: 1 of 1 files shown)


nknapp:
Hi and thanks for reporting this. I think it would be better to include a whitelist of files into the package.json's `files`-section (see https://docs.npmjs.com/files/package.json#files for details). The `.npmignore`-file would be obsolete then. Could you do that?

Also, currently, in order to have your change affect the next release, you should point it to the 4.x branch. I will then merge everything into the master.

aaharu:
thanks for reviewing.

I agree your opinion :+1:
I will change it within a few days.

PR #1408, opened 2017-11-25
Use `files` field instead of `.npmignore`

aaharu:
- [ ] Have tests

referenced #1407

These are removed files from package.
```diff
Only in node_modules/handlebars: .gitattributes
Only in node_modules/handlebars: .gitmodules
Only in node_modules/handlebars: .idea
Only in node_modules/handlebars: .istanbul.yml
Only in node_modules/handlebars: .npmignore
Only in node_modules/handlebars: CONTRIBUTING.md
Only in node_modules/handlebars: FAQ.md
Only in node_modules/handlebars: appveyor.yml
Only in node_modules/handlebars: docs
Only in node_modules/handlebars: package-lock.json
```


package.json
```diff
@@ -73,5 +73,15 @@
     "buildConfig": {
       "minify": true
     }
-  }
+  },
+  "files": [
+    "bin",
+    "dist/*.js",
```
nknapp: I think you could replace all the `dist/...`-entries with just one `dist` entry.

nknapp:
I've added a small comment to the commit. I think one `dist` entry is sufficient. The rest looks good.
You don't need to open another PR for that change. Just push or amend and force-push. 
I'll squash-merge later if needed.

nknapp:
I think I have forgot about the components... I'll have a look tonight, but I might just take your first commit after all.  Thanks.

nknapp:
I have merged your first commit manually.

nknapp:
Is it important that you have a release with this change anytime soon? Otherwise, I would wait for other changes.

aaharu:
thank you.
It is not important for me 😸

PR #1415, opened 2017-12-28
Update README.markdown

jfoclpf:
it might be obvious for you, but for the sake of clarity it should be mentioned how such `Handlebars` variable appears

- [x] Have tests

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -28,6 +28,9 @@ the template into a function. The generated function takes a context
 argument, which will be used to render the template.
 
 ```js
+
+const Handlebars = require('handlebars');
+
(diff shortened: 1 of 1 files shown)


nknapp:
Your change is correct if you use a common-js module loader. But it differs from system to system. If you are not using node, browserify or webpack, but the umd-bundle, then you don't need this line...

I admit that this is not for production, but then you wouldn't use #compile() in the browser in production anyway.

nknapp:
Maybe a comment would be appropriate. A line that states that handlebars must be loaded globally or through your favourite module loader

jfoclpf:
Ok. Thank you for the clarification

Issue #1386, opened 2017-09-26
Request to support babel 6


pravi:
We'd like to use babel 6 to build handlebars in debian. Currently build fails with this error
```
Warning: [BABEL] lib/handlebars.js: Using removed Babel 5 option: base.loose - Specify the `loose` option for the relevant plugin you are using or use a preset that sets the option. Use --force to continue.
```

nknapp:
There is a work-in-progress PR at #1271. If you want to pick it up and finish it, you are welcome.

nknapp:
Please make an PR against the 4.x-branch, because the master will not be release in the near future.

nknapp:
I've made an attempt in the branch "babel6". The webpack-configuration is still missing, and I'm not sure this works with webpack 1.x. Maybe it needs to be upgraded as well, but I don't know if I have the time for that now.

If anyone wants to pick this up (on my branch or on #1271, go ahead).

nknapp:
If anyone feels inclined to work on this problem, please check out the branch `4.x-babel6` and create PRs to this branch.

nknapp:
@pravi I don't fully understand why you can't use babel5 to compile handlebars. Do you have to use the  version that comes packaged with debian?

This update is harder than it looks because there the breaking changes in the module loading behavior in babel@6 (see the discussion at #1402 and the changelog at https://github.com/babel/babel/blob/master/CHANGELOG.md#600 

The changes look valid and they are actually bugfixes, but they are not backward compatible and cause the handlebars-tests to fail. 

This can only be pushed into Handlebars 4.x, if

* all tests pass
* the interface stays the same: 
  * especially the `var Handlebars = require('handlebars')` for nodejs-users
  * and also the the syntax when using es6-imports, but I guess this depends more on the babel-version that the importing library uses.

Otherwise this would have to be Handlebars@5 and I don't see that happen, unless sombody starts working on a formal specification of the language (#1277).

As fas a I'm concerned. I am currently *not* working on this issue, but I will look into pull-requests and discussions.

pravi:
@nknapp yes, I have to use the version packaged in debian, which is babel 6. Usually I can manage to update to a newer version (like if handlebars used babel@7, I can update babel in debian to 7, but moving to old versions are practically discouraged, but possible only in extreme cases. This also means maintaining multiple versions of babel, which we always try to avoid.)

pravi:
[...]

PR #1416, opened 2018-01-01
Webpack upgrade

jishnu7:
Pull request to upgrade webpack and babel to respective latest versions. All test cases are working. 

There is an issue with coverage. Istanbul doesn't support ES6-Babel properly, we will have to use either [babel-istanbul](https://github.com/jmcriffey/babel-istanbul) or [isparta](https://github.com/douglasduteil/isparta). Since master branch is using a custom fork of istanbul, I'm letting you decide on how to proceed with it. 

Another way to handle is, running test cases on original source rather than the babel generated one. If we are going to do that, then we don't need grunt-babel, we will be able to use babel-loader with webpack to generate just the dist files we need, rather than exporting all transpiled sources.

Relates to: #1386 ,  #1399 ,

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,3 +1,5 @@
+const path = require('path');
+
 /* eslint-disable no-process-env */
 module.exports = function(grunt) {
 
@@ -46,12 +48,16 @@ module.exports = function(grunt) {
     babel: {
       options: {
         sourceMaps: 'inline',
-        loose: ['es6.modules'],
+        presets: [
+          ['env', { loose: true }]
+        ],
         auxiliaryCommentBefore: 'istanbul ignore next'
       },
       amd: {
         options: {
-          modules: 'amd'
+          presets: [
+            ['env', { modules: 'amd' }]
+          ]
         },
         files: [{
           expand: true,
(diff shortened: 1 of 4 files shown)


nknapp:
That sounds great. [remap-istanbul](https://github.com/SitePen/remap-istanbul) would be another option. I have used it in a TypeScript-project and it worked fine. It needs node 6 though, but that is another discussion.

In any case, could you point your PR to the 4.x-branch? That is the branch of currently released versions. There has not been a release of the master branch yet. I'll merge into the master from 4.x then.

4.x is not using the custom fork of istanbul, so `babel-istanbul` might work fine.

jishnu7:
Made necessary changes and target branch to v4.x. 

Using babel-istanbul straight away cleared the coverage except branches, which is having 99.91% coverage. Only one branch is not covered, but the interface is not clearly pointing out which section of the file is causing the issue. I'm checking if we can do something about it.

jishnu7:
amd build seems to be failing because of dynamic import. master branch doesn't seems to have amd builds. Do you use amd build anywhere?

nknapp:
I don't know how many users Handlebars has, but I think the pobability that *someone* uses amd-builds, is pretty high.

nknapp:
I'll close this for now. Thanks for trying back then. I think, with some more integration tests we can do another attempt, but not now.

Issue #1390, opened 2017-10-11
When precompiling templates using the --amd --handlebarPath flags the path is modified internally before output


SamFromDaUk:
I am using requirejs to compile my project for production, however I am running into an issue compiling templates.

I have an require.js build file (simplified)

```
({
    baseUrl: '../build/js',
    paths: {
        'handlebars': '../../../node_modules/handlebars/dist/handlebars.runtime.min'
    },
    out: '../build/js/config.js',
    include: ['index']
})
```

By using requirejs in this way, I need to define the path to handlebars in an AMD module as 

```define(['handlebars'], function(Handlebars) {```

However due to https://github.com/wycats/handlebars.js/blob/master/lib/precompiler.js#L175, this is modified to the below example. (see the string concatenation to add `handlebars.runtime`:

```define(['handlebarshandlebars.runtime'], function(Handlebars) {```

I cant see a clean way to remove `handlebars.runtime` from the output of the precompiler.

I would propose that the user of the cli be able to specify the full path to the handlebars file, however I imagine this would be a breaking change.

What would you suggest?

SamFromDaUk:
Hi, does anyone have any feedback on this issue? The change is quite minor (and I would be happy to make the PR myself), but I believe this may break other use cases.

nknapp:
Hi. I think it would be OK to have an additional option to the precompiler, maybe "runtimeModuleName", that defaults to the current value "handlebars.runtime".

If you want to open a PR, please do so against the 4.x branch

nknapp:
Closing due to inactivity

PR #1417, opened 2018-01-03
#1390 Added runtimeModuleName cli variable, defaults to the previous value of handlebars.runtime

SamFromDaUk:
See https://github.com/wycats/handlebars.js/issues/1390.

I haven't added tests yet for this change as I could not see any existing ones for `handlebarPath` or `namespace`. Please can you advise if you would like these adding, and where I should do this?

Is there a separate repository for documentation I should update for this new CLI variable?

### DIFF
--- a/bin/handlebars
+++ b/bin/handlebars
@@ -24,7 +24,7 @@ var optimist = require('optimist')
       },
       'h': {
         'type': 'string',
-        'description': 'Path to handlebar.js (only valid for amd-style)',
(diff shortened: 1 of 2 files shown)


nknapp:
Here is a test with `namespace` as option: https://github.com/wycats/handlebars.js/blob/master/spec/precompiler.js#L138-L144

nknapp:
We can reopen this PR once there is a test. I will close it for now in order to clean up the PR list.

PR #1467, opened 2018-09-11
Remove duplicate text in release-notes.md

sharang-d:
Hello, I was reading through the release notes in order to upgrade the version of the handlebars package I was using and found that the release notes for v4.0.12 were duplicated twice. I've removed the duplicate text blocks. 

Thanks!


nknapp:
Oops

PR #1498, opened 2019-02-07
show source location for the strict lookup exceptions

sohibegit:

- [x] Have tests
- [ ] Don't significantly decrease the current code coverage (see coverage/lcov-report/index.html)

### DIFF
--- a/lib/handlebars/runtime.js
+++ b/lib/handlebars/runtime.js
@@ -67,9 +67,9 @@ export function template(templateSpec, env) {
 
   // Just add water
   let container = {
-    strict: function(obj, name) {
-      if (!(name in obj)) {
-        throw new Exception('"' + name + '" not defined in ' + obj);
+    strict: function(obj, name, loc) {
+      if (!obj || !(name in obj)) {
+        throw new Exception('"' + name + '" not defined in ' + obj, { loc: loc });
       }
       return obj[name];
     },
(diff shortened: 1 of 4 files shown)


wycats:
@sohibegit would you be willing to add a test for this?

sohibegit:
@wycats Sure I'll do it when I have time.

wycats:
@sohibegit Do you think you'll have time to add a test?

sohibegit:
@wycats done.

lib/handlebars/runtime.js
```diff
@@ -67,9 +67,9 @@ export function template(templateSpec, env) {
 
   // Just add water
   let container = {
-    strict: function(obj, name) {
-      if (!(name in obj)) {
-        throw new Exception('"' + name + '" not defined in ' + obj);
+    strict: function(obj, name, loc) {
+      if (!obj || !(name in obj)) {
```
nknapp: Could you explain, why you have added this part (`!obj || ...`).

nknapp:
I'm sorry to say this, but I haven't noticed before: This PR targets the master branch. Development of the 4.x version, currently happens in the 4.x branch.

Could you replace your commits onto 4.x an update the PR? 

I have noticed one odd thing that I will add in a review.

lib/handlebars/runtime.js
sohibegit: because of this case `{{foo.bar}}` in strict when foo is unknown it will throw `cannot read property 'bar' of undefined `before it can throw the exception with source location.

sohibegit:
the new pr is found here #1527

Issue #1492, opened 2019-01-02
3.1.x: Option to configure escaping in Handelbars.escapeExpression (escape "=")


nknapp:
This is a follow-up issue for #1454. It's goal is to discuss the implementation of a fix for [security advisory 61](https://www.npmjs.com/advisories/61) in the 3.x branch.

Some people have asked to fix the issue in 3.0.x (#1454). When this was done, other people rightfully claimed that this was a semver-violation (#1489). The fix was reverted in 3.0.6, but we still need an option
to resolve the problem.

The only way to comply to semver rules is to build a 3.1 version with an option to activate escaping of "=". @wycats also proposed to add a warning output if the option is not set.

The most simple way would be:
- [ ]  Add a function `Handlebars.escapeEqualSign()` that replaces `escapeExpression` in the Handlebars environment by a version that escapes the "="-character as well.
- [ ] Print a warning when `escapeExpression` is called and the property is not set.

The drawback would be that the warning would only be printed in unit-tests and on the production site, which would invite attackers to exploit the vulnaribility.

Another way would be to:

- [ ] Add an optional second argument (an `options`-object) to `Handlebars.escapeExpression` that activates escaping of "=".
- [ ] Add a compiler option that adds the ensures that `Handlebars.escapeExpression` is always called with this option, when `{{mustache}}` expressions are found in the template.
- [ ] Print a warning, if the compiler option is not set.
- [ ] Print a warning, when `escapeExpression` is called *for the first time* without the option (no need to be overly verbose)
- [ ] Optional: Add a confguration method `Handlebars.failIfUnsafeEscape()` that causes `Handlebars.escapeExpression` to throw an exception when called without the option. This could be used in builds to find all the places where the option is still missing.

This would allow developers to add the option to their precompile step and the warning would usually only be printed at build-time and not presented to the users (and potential XSS-hackers).

Either way, the version 3.1 would still be marked as insecure in the security-database. 

Since I am not personally using this version anymore, I'd like to hear preferences or other ideas from people who need this change.

/cc @mattolson @courtneynguyen

nknapp:
Closing this due to inactivity, please comment or submit a PR if you want to see this in the 3.x branch

Issue #1489, opened 2018-12-20
Handlebar to html conversion broken in 3.0.5


rhariraman:
<select class={{classes}} {{attrs}}>
  {{#options}}
    <option id="{{id}}" value={{value}} {{defaultSelected}}> {{label}} </option>
  {{/options}}
</select>

In the above handlebar file we have replaced the attribute with the following code a long back and it was working as expected,
attrs: "data=#{dataContext.id.toString()} style=width:70px;"

This was perfectly working in the 3.0.3 version and properly interpreted to the following code,

data="5336686" style="width:70px;"

But when we upgraded to 3.0.5 the above code is breaking and giving the following encoded code output,

`data&#x3d;5339575 style&#x3d;width:70px;`

As far as my understanding, all the minor and patch version releases are backward compatible. I would like to know that we should revamp the above code or this could be reverted in the next minor release? This has caused a major production issue for us :( Thanks in advance!

huijar:
We have the exact same issue. In our case we have base64 encoded strings which are otherwise left intact, but any trailing `=` characters get converted to `&#x3D;`.

jlaari:
+1, exactly same issue here.

nknapp:
Hi. I'm sorry, if this change causes problems for you. It was introduced via #1083. I have published it in response to #1454, which references an entry in the  [node security database](https://nodesecurity.io/advisories/61).

But you are right. The list of escaped characters was documented, so this was a breaking  change. 
I'm going to revert it, but I will also get the oppinions of the people who wished to have the change.

@rhariraman By the way: You are using the lack of "="-escaping exactly in the way that may cause the vulerability (see https://github.com/wycats/handlebars.js/pull/1083#issuecomment-134733449)

nknapp:
The security fix is reverted in version 3.0.6

rhariraman:
@nknapp Thanks a lot :)

Ravikant-M1044571:
> any trailing `=` characters get converted to `&#x3D;`

facing same issue with version '4.2.0'. what is the solution for this plz.

chrisft25:
> > any trailing `=` characters get converted to `&#x3D;`
> 
> facing same issue with version '4.2.0'. what is the solution for this plz.

You could use triple-stash to avoid HTML conversion. Example: {{{text}}}

PR #1514, opened 2019-03-18
V1.3.x (Retrofit Security Patch...)

infolock:
granted - this will probably be closed as soon as it is opened - however... this _would_ help teams that may have a legacy build that still uses this old ass version ;)

Original PR this PR is based on can be found here: https://github.com/wycats/handlebars.js/commit/83b8e846a3569bd366cf0b6bdc1e4604d1a2077e

Idea is simply to have a `1.3.x` branch so that a `1.3.whatever` tag can be generated from it with the patch applied...

if you guys could create a `v1.3.x` branch and update this PR to be merged into it - that would effectively do what we're after here...

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -96,7 +96,7 @@ module.exports = function(grunt) {
     uglify: {
       options: {
         mangle: true,
-        compress: true,
+        compress: {},
         preserveComments: 'some'
       },
       dist: {
@@ -169,10 +169,12 @@ module.exports = function(grunt) {
 
   // Build a new version of the library
   this.registerTask('build', "Builds a distributable version of the current project", [
+                    'clean',
+                    'jshint',
                     'parser',
                     'node',
-                    'globals',
-                    'jshint']);
+                    'globals'
+                  ]);
 
   this.registerTask('amd', ['packager:amd', 'requirejs']);
   this.registerTask('node', ['packager:cjs']);
--- a/lib/handlebars/compiler/javascript-compiler.js
+++ b/lib/handlebars/compiler/javascript-compiler.js
@@ -11,6 +11,10 @@ JavaScriptCompiler.prototype = {
   // PUBLIC API: You can override these methods in a subclass to provide
   // alternative compiled forms for name lookup and buffering semantics
   nameLookup: function(parent, name /* , type*/) {
+    if (name === 'constructor') {
+      return ['(', parent, '.propertyIsEnumerable(\'constructor\') ? ', parent, '.constructor : undefined', ')'];
+    }
+
     var wrap,
         ret;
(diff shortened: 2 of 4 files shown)


nknapp:
Yes, your probbably right, it won't be merged. I have attempted to fix that in the 3.0.x branch. Shortly after publishing 3.0.4 and 3.0.5 with the fix, I received the first complaints (#1489) that this breaking change should not be done with an patch-version increment. But as I wrote in #1492: The only semver-compatible way to achieve this, is a compiler- or runtime-option that allows you to activate secure escaping. It's definitely not a patch increment or a `1.3.0-something`

But I have created a `1.x`-branch.

infolock:
Note: I intentionally left the PR in a conflict state until the wycats team can have an opportunity to decide if they even _want_ to do this...

infolock:
@nknapp heh, i'm not entirely sure why the thing this breaks would actually be considered a reason not to apply this (isn't it doing exactly what it was intended to do? 🤔 )

Either way, this PR was creating just as a means to help get it applied to that build (if it is even possible ;)

infolock:
i dunno - the person's complaint about it breaking is using it in the way that was outlined in the original report : "When using attributes without quotes in a handlebars template, an attacker can manipulate the input to introduce additional attributes, potentially executing code. " - [source](https://snyk.io/vuln/npm:handlebars:20151207)

so i'm not so sure this should have been reverted.   i'll watch the threads related to this though and see where you all go from here.

imo though - this shouldn't be considered a handlebars vulnerability so much as it is a vulnerability introduced by the `developer(s)` who aren't applying quotes to values.  but hey - that's just me ;)

infolock:
@nknapp  Actually - the more I'm reading [the reported issue on snyk](https://snyk.io/vuln/npm:handlebars:20151207), the more i'm now convinced this is not a handlebars issue at all.   

i'm going to close this PR because there isn't actually anything to fix in handlebars.   the developer who reported this issue needs to just fix their own code lol...

sorry for all the @ mentions - just wanted to keep you informed here...

infolock:
@nknapp i actually thought maybe handlebars wasn't responsible for this and [hoped to get the vulnerability report removed entirely](https://github.com/nodejs/security-advisories/issues/12).  however (and after a few days of back and forth discussions) it turns out that it unfortunately isn't that simple.

However... is there any way that the escape character logic and be removed completely and the documentation updated to allow users understand that inserting HTML without quotes around their attributes would result in an XSS injection, and that handlebars does not condone nor attempt to protect against this (or something along those lines)?

If you do - then this would satisfy the report and this can be lifted.

Otherwise, I'm afraid you are in for a very long rabbit hole of possible bad characters that could be used as a means to create an XSS Vulnerability.

PR #1515, opened 2019-03-19
Port over linting and test for typings

zimmi88:
- Move typings from lib/ to types/.
- Add dtslint for validating types.
- Use grunt-bg-shell to call out to dtslint during build step.

Notes:
- `dtslint` is the tooling that was used previously to check types, and it's designed to validate typings, so it seemed like a good fit for this use case. Much of the directory structure in this PR mimics the instructions from https://github.com/Microsoft/dtslint.
- Couldn't find a grunt library to run `dtslint`, so I'm using `grunt-bg-shell` to proxy to a script in `package.json` that runs `dtslint`.
- Originally, I ran into an issue where `dtslint` was complaining about types in `node_modules` which seemed odd. But, after making a couple of tweaks to the typings test file, I'm not able to reproduce this anymore. I'm relying on the test run against this PR to tell me if this was a fluke or if the issue persists.

=========

- [x] Have tests


nknapp:
I think I would rather create a custom grunt task in the tasks-directory, than including another pacjage. Just use child_process.execFileSync to run dtslint.

But we can also merge your commit first, and then change the task. I'll look into it.

Thanks for supplying this.

nknapp:
OK, never mind. I feared that grunt-bg-shell would be pulling tons of dependencies into tree. But it doesn't have any, so we might as well keep it.

nknapp:
Released in 4.1.2

Issue #858, opened 2014-09-03
Disable new default auto-indent at included partials


majodev:
Handlebars >v2.0.0-alpha.4 introduced a new default indentation handling at partials, namely "_Partials that are standalone will now indent their rendered content_" ([source](https://github.com/wycats/handlebars.js/blob/master/release-notes.md#v200-beta1---august-26th-2014)). This might cause unexpected behavior when upgrading from previous versions, is there a `flag` to restore the old handling?

[Detailed explanation / self reproduce](https://github.com/majodev/handlebars-2-0-default-indentation-handling)

Thanks in advance! :)

lzilioli:
This same behavior appears to exist when calling partials, or outputting content with the `{{{ }}}` syntax. Is there a way to prevent this behavior in these cases as well?

kpdecker:
@lzilioli please open a new issue with supporting fiddle or other repo case if you're running into issues. I'm not sure what you mean by the same behavior as there are a few things discussed here, quite awhile ago, and it's probably better to start a new standalone conversation if you have concerns.

erik-landvall:
Is there still no fix for this implemented? This renders the complete library useless for a lot of users, just to have a bit pettier output..? Until a working fix for the prettier output can be produced, the feature should be reversed, as it is broken.

PR #1518, opened 2019-03-29
Prevents indent on blank lines of partials.

bdiz:
This pull request is to prevent indent in partials that have blank lines. It is needed when handlebars is used as a code generator and the spacing of the output is important or linted.

Here is the result of the test in the pull request before the fix:
image

The blank line in Yehuda's function has indent. This is not desirable. After the fix, this line is no longer indented.

- [x] Have tests


nknapp:
Sounds like a breaking change to me, and a pretty special use-case. It should at least be a feature switch. 

Even then... Couldn't you just add a post-processing step like 
```
const withoutEmptySpaces = string.replace(/^\s+$/gm,"")
```

or use `prettier` to format the result? Seems easier to me than changing Handlebars.

bdiz:
Thanks for taking a look.

Here is some output which is bit more illustrative of our use case.

image

We really see this as a bug in the indentation algorithm. We feel it's not clean code to add indent on the blank line of the function above. Anybody generating code that will be linted would benefit from this fix. 

What use case would want indent for blank lines of a partial?

We could post-process, as you recommend, but we use this extensively at https://www.uvmgen.com and since we are reactively generating code on the front end, performance is important to us. This does not add any performance overhead to handlebars since it already has a condition checking for blank lines as it adds indent.

I'm not understanding this being a breaking change.

Adding indent originally (https://github.com/wycats/handlebars.js/issues/858) needed the `preventDefault` feature switch because indentation was a new feature. If you consider this a bug fix as we do, a feature switch may not be needed.

nknapp:
I have become very cautious with changes and very cautious with the assumption that "nobody would use Handlebars this way". The lack of a formal language specification makes it difficult to say what is a bug and what isn't. The large number of users makes it probable that *one* of them relies on exactly that behavior that you want to change, even if it seems insignificant to you and me.

You are right that when using Handlebars to create JavaScript, Java or alike, it seems natural to skip indent on blank lines. In Python, the majority also seems to prefer blank lines without indent. But there are some that say blank lines should be indented (https://stackoverflow.com/a/25002433).

Without any target file type in mind, treating blank lines and non-blank lines the same seems natural to me, so I don't consider this a bug.

I cannot say for sure that your change does or doesn't break any build, except by publishing a new version with the change. Then, if just one user complains, I will have to revert the change and publish a new version. The user may have had an hour of debugging until she finds out that a change in Handlebars is responsible for her test failures. And I have spent at least an hour publishing two patch versions (publishing is not completely automated yet).

But since this affects that language, I'd like to get @wycats opinion, too.

nknapp:
Oh, it just came to my mind that if this is *not* a bug, than it is a new feature that changes the language. I promised @wycats not to allow any new features, until there is a formal language specification.

Another reason to delegate this decision to @wycats

wycats:
@nknapp thanks for being cautious!

I tend to think that this is on the border between a feature and a bug, but you're right that it's very likely to end up being a breaking change.

I wouldn't be opposed to collecting use-cases for a new revision of the whitespace stripping algorithm (indentation sensitivity comes to mind as a justification for a revision), and in that context, I think this change would make sense.

auvipy (approved):

mohd-akram:
IMO this would be a good change for 5.x.

Issue #1277, opened 2016-11-23
We need a spec


lawnsea:
Currently, the Handlebars language is defined by the behavior of this reference implementation and incompletely described by the documentation found at handlebarsjs.com. This makes it very difficult to reason about bugs like #1276. Should partials that use `{{> @partial-block}}` be invocable in the manner the bug reporter expects? Not clear.

I had a related discussion with @wycats and we came to the conclusion that we need to specify the language before we move forward with any new features.

This would be a non-trivial effort. Thoughts?

lawnsea:
#1262 is another great example of why we need a spec.

lawnsea:
Looks like work is already underway: https://github.com/handlebars-lang/spec.

nknapp:
I'm reopening this issue. Although there is an attempt for a spec in the referenced repo, there is no progress at the moment.

MaxLeiter:
No progress has been made in months, and no new features are being added until it's done? That seems incredibly... silly

nknapp:
I think, we need more help here. Someone has made a first attempt for a spec, but at the moment, there is nobody actively working on it. On the other hand @wycats has made it clear, that he wants to have a spec before new language features are implemented. The reason for this is:

There are many implementations of the Handlebars language in different programming-languages (Java, Rust, multiple implementations in JavaScript...) and without a formal specification, they will always diverge. You can never be sure that your templates still work when you migrate to another engine.

nknapp:
Yes, there aren't many active collaborators at the moment. I don't want to say that I'm the only one, but sometimes it feels like it.

When I asked @wycats for push-access and release-permissions, I had the feeling that there was noone with the time and the resources to fix bugs and publish new releases, so I wanted to at least become that person.

I would also love to write a spec and a great documentation site, but the reality is that I have a wife and two kids, a daytime job and a house to pay-off. And I have few other open-source projects.

I can fix bugs, answer to issues, merge PRs and publish new versions. But writing a spec is too much (at the moment)

thislooksfun:
Has there been any progress on this? I'm attempting to write a Dart version of handlebars, and it is very difficult to without a formal spec to work off of.

nknapp:
I have started a repo at handlebars-lang/specification on which I work when I have nothing else to do (which is not very often). Feel free to contribute there. I can add you to the team if you want.

PR #1523, opened 2019-04-17
feat: allow a custom escapeExpression for an env

thekiwi:
We use handlebars.js for both XML templating and JSON templating within the same application. As such, we'd like to be able to override the escapeExpressions function on a single Handlebars environment, rather than globally. This is a small PR that allows us to do that, by first attempting to get the environment's version of `escapeExpression` and falling back to the `Utils` version if that is falsey.

Usage (or how we might end up using it):
```JavaScript
const hb = Handlebars.create();
hb.escapeExpression = str => JSON.stringify(str).slice(1, -1);
// ... continue using 'hb' as normal
```

Please let me know if anything is unclear or if you'd like me to make additional changes.

- [x] Have tests

### DIFF
--- a/lib/handlebars/runtime.js
+++ b/lib/handlebars/runtime.js
@@ -88,7 +88,7 @@ export function template(templateSpec, env) {
       return typeof current === 'function' ? current.call(context) : current;
     },
 
-    escapeExpression: Utils.escapeExpression,
+    escapeExpression: env.escapeExpression || Utils.escapeExpression,
     invokePartial: invokePartialWrapper,
 
     fn: function(i) {
--- a/spec/basic.js
+++ b/spec/basic.js
@@ -122,6 +122,16 @@ describe('basic context', function() {
           'escaping should properly handle amperstands');
   });
 
+  it('escaping with escapeExpressions override', function() {
+    global.handlebarsEnv.escapeExpression = s => s.replace('&', 'AND');
+    shouldCompileTo('{{awesome}}', {awesome: 'x&y'}, 'xANDy', 'text is escaped using the custom escapeExpression function');
+  });
+
+  it('falls back to the default escapeExpressions', function() {
+    global.handlebarsEnv.escapeExpression = null;
+    shouldCompileTo('{{awesome}}', {awesome: 'x&y'}, 'x&amp;y', 'text is escaped using the standard escapeExpression function');
+  });
+
   it("functions returning safestrings shouldn't be escaped", function() {
     var hash = {awesome: function() { return new Handlebars.SafeString('&\'\\<>'); }};
     shouldCompileTo('{{awesome}}', hash, '&\'\\<>',


thekiwi:
Hey @nknapp, just wondering whether you had a chance to look at this? Is it a small enough addition that you would be happy to merge and release? Thanks!

nknapp:
Sorry, I haven't had a thorough look yet. But I think something similar is part of our plans avoid security vulnerabilities in the future. I would like to defer until then to avoid unnecessary api changes.

nknapp:
I would like to merge this, but the API is not what I would want to have. I have to do a general API discusssion with @wycats for overriding this like that (helperMissing, blockHelperMissing, escapeExpression). I don't know when I will be able to do that and it might fall in the ["we need a spec"](#1277) category.

I would think of an interface similar to

```js
Handlebars.registerHook('escapeExpression', (string) => ...);
// or 
Handlebars.hooks.escapeExpression((string) => ...)
```
Technically, this is not a language feature, so I'm not sure ich 'we need a spec' applies here, but in the end, that @wycats decision.

thekiwi:
@nknapp I'd be very happy with your proposed change. I made my PR change as small as possible to minimise the impact, but totally understand if you'd like to implement this as a proper supported and documented API.

nknapp:
In #1559, I added the `hooks` property to the runtime. I'm still not confident about a good name for the function(s) to register a hook

* `Handlebars.on('escapeExpression', (expression) => ...))` would match the common code from Node.js when it comes to events, but this is not an event.
* `Handlebars.registerHook('escapeExpression', (expression) => ...))` would match the current helper-registration in a way. But it bugs me that there is a string defining the hook, because there is no arbitrary set of hooks to be replaced.
* `Handlebars.hooks.escapeExpression((expression) => ...))` is not possible at the moment, because the `hooks` object already contains hooks, but actually, the `hooks` should be private, so maybe we should call them `_hooks` anyway. Still, the method name does not suggest that the escapeExpression-function is being replaced.
* `Handlebars.replace.escapeExpression((expression) => ...))` would be more descriptive. But it feels weird having a `replace` property. And it would sound wrong for `helperMissing`.
* `Handlebars.hookInto.escapeExpression()` ...

I don't know. Does anyone have a good idea, or knows how other frameworks solve this?

nknapp:
@thekiwi  would it be OK for you to have a runtime option to override "escapeExpression"?

thekiwi:
@nknapp It should be, are you able to provide a snippet demonstrating how I would use it?

nknapp:
It would be something like 

```js
const template = Handlebars.compile('{{name}}');
const output = template({ name: 'Jonny "the hound" Walker' }, {
  hooks: {
     escapeExpression: string => string.replace(/(["'\\])/g,'\\$1')
  }
}
```

Maybe wrapped in a 'hooks`

KrishnaPG (approved):
This is better than having to override the global utils.escapeExpression.

Till some other mechanism (such as hooks() etc.) is made available, this should be strongly considered for merging.

silverwind:
Ideally I too would like to see a interface that does not require the use of `Handlebars.create`, but I tend to agree this is mergeable as-is, just to unblock a number of people who would like to use Handlebars for non-HTML templating with proper escaping.

silverwind (approved):

PR #1527, opened 2019-05-29
show source current location for the strict lookup exceptions

sohibegit:

- [ ] Please don't start pull requests for security issues. Instead, file a report at https://www.npmjs.com/advisories/report?package=handlebars
- [ ] Maintain the existing code style
- [ ] Are focused on a single change (i.e. avoid large refactoring or style adjustments in untouched code if not the primary goal of the pull request)
- [ ] Have tests
- [ ] Don't significantly decrease the current code coverage (see coverage/lcov-report/index.html)
- [ ] Currently, the `4.x`-branch contains the latest version. Please target that branch in the PR.

### DIFF
--- a/lib/handlebars/compiler/javascript-compiler.js
+++ b/lib/handlebars/compiler/javascript-compiler.js
@@ -1072,6 +1072,7 @@ JavaScriptCompiler.prototype = {
 
   setupHelperArgs: function(helper, paramSize, params, useRegister) {
     let options = this.setupParams(helper, paramSize, params);
+    options.loc = JSON.stringify(this.source.currentLocation);
     options = this.objectLiteral(options);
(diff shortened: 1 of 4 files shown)


lib/handlebars/compiler/javascript-compiler.js
```diff
@@ -1072,6 +1072,7 @@ JavaScriptCompiler.prototype = {
 
   setupHelperArgs: function(helper, paramSize, params, useRegister) {
     let options = this.setupParams(helper, paramSize, params);
+    options.loc = JSON.stringify(this.source.currentLocation);
```
sohibegit: This is so I can be able to access the source location from inside helpers it's useful when I want to throw an exception including the location from inside my helper

PR #1534, opened 2019-07-03
Add VM TS mapping

AndrewLeedham:
I am aware VM is not part of the public API. But, In one of our projects we are overridding the `resolvePartials` function in order to do some custom resolution, it would therefore be useful if it was mapped to TS to aid in that.

Not sure if these typings are 100% accurate, I'm not particularly familiar with the structure of handlebars.js behind the scenes, but I took my best stab at it.

- [x] Have tests

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -5,15 +5,15 @@ module.exports = function(grunt) {
     pkg: grunt.file.readJSON('package.json'),
 
     eslint: {
-      options: {
-      },
       files: [
         '*.js',
         'bench/**/*.js',
         'tasks/**/*.js',
         'lib/**/!(*.min|parser).js',
         'spec/**/!(*.amd|json2|require).js',
-        'integration-testing/**/*.js'
+        'integration-testing/multi-nodejs-test/*.js',
+        'integration-testing/webpack-test/*.js',
+        'integration-testing/webpack-test/src/*.js'
       ]
     },
 
--- a/lib/handlebars/runtime.js
+++ b/lib/handlebars/runtime.js
@@ -209,6 +209,9 @@ export function wrapProgram(container, i, fn, data, declaredBlockParams, blockPa
   return prog;
 }
 
+/**
+ * This is currently part of the official API, therefore implementation details should not be changed.
+ */
 export function resolvePartial(partial, context, options) {
   if (!partial) {
     if (options.name === '@partial-block') {
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -8,6 +8,7 @@
  *   - Sergei Dorogin <https://github.com/evil-shrike>
  *   - webbiesdk <https://github.com/webbiesdk>
  *   - Andrew Leedham <https://github.com/AndrewLeedham>
+ *   - Nils Knappmeier <https://github.com/nknapp>
  * For full history prior to their migration to handlebars.js, please see:
  * https://github.com/DefinitelyTyped/DefinitelyTyped/commits/1ce60bdc07f10e0b076778c6c953271c072bc894/types/handlebars/index.d.ts
  */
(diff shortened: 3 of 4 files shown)


AndrewLeedham:
@nknapp Does this seem right?

AndrewLeedham:
@nknapp would be useful to know if this is likely to get merged? Perhaps the typings should be added back to DefinitelyTyped?

@zimmi88 what was the original reason for including the typings here?

nknapp:
I missed this one, sorry about that...

Honestly, I'm not sure if I want to support this. Adding those type definitions would make those internals an official API which means that we would not be able to change it without a major version increase.

On the other hand, the `resolvePartials`-method is mention in the [compiler-api](https://github.com/wycats/handlebars.js/blob/master/docs/compiler-api.md#javascript-compiler) so it is already kind of official.

The compromise I would be willing to accept is: 

* Add `VM` with just the `resolvePartials` function and omit the other functions.
* Add a `/** @deprecated */` annotation to the function to clarify that this is not officially endorsed anymore.
* Add a test to `types/test.ts`
* Add a comment to the implementation of the function, warning that it is currently
  part of the official API and should not be changed.
* open a new issue to discuss a new way for providing the functionality.

AndrewLeedham:
@nknapp Makes sense, I have implemented your suggestions and opened a discussion issue: https://github.com/wycats/handlebars.js/issues/1545

nknapp:
Thank you. I have made some adjustments. The "options" object of "resolvePartial" seems to be different from the RuntimeOptions. Some options are missing and the "name" is new. I have used the debugger to inspect the object and decided that it should be its own type.

I hope it was ok to force push on your branch. I neede to do that after squashing and rebasing the branch.

nknapp:
Released in 4.2.0

AndrewLeedham:
> Thank you. I have made some adjustments. The "options" object of "resolvePartial" seems to be different from the RuntimeOptions. Some options are missing and the "name" is new. I have used the debugger to inspect the object and decided that it should be its own type.
> 
> I hope it was ok to force push on your branch. I neede to do that after squashing and rebasing the branch.

That is fine, makes more sense now. Thanks for getting this in :)

AndrewLeedham:
Actually @nknapp it seems that `name` does exist outside resolvePartials. See https://jsfiddle.net/AndrewLeedham/jkpb3equ/6/, it exists during `compile` if it is currently compiling a partial. So perhaps it should be optional on the `RuntimeOptions` interface?

nknapp:
The "name" is indeed set to the runtime options, when a partial is called. But I would say this is a collateral effect, not intended. Passing "name" to a compiled template does not have any effect, in the template as far as I can see.

Do you need the "name" to be in there?

AndrewLeedham:
> The "name" is indeed set to the runtime options, when a partial is called. But I would say this is a collateral effect, not intended. Passing "name" to a compiled template does not have any effect, in the template as far as I can see.
> 
> Do you need the "name" to be in there?

We are currently using it yes. It is an unusual use case, but we are checking the name against a map of Mustache templates and rendering it if it exists, otherwise just continue the normal handlebars render.

nknapp:
I can see *how* you are using it? Where do you insert the logic that checks the name against the template map. Are you overriding the "compile" function? 

I'm asking because I don't think that this should be in the official API. Overriding any functions in the Handlebars-prototype is not intended and should not be promoted. Every addition to the API means that there is one more case where I would have to increase the major version of the library. But I would like to maintain compatibility, because that is least hassle for all users.

You can send me a short fiddle that shows how you are doing it. I may change my mind and add it. Please create another issue for that, so that it is better to find.

But after reading "Clean Code" by Robert C. Martin, and looking at the Handlebars source code, I feel the urge to do some refactorings. The code involving the "name" option. May also be subject to change in the future.

If you give me a usage example, I might keep it in mind when changing that code.

In any case, you can still add you own type definitions in your own projects, merging the "name" into the options.

https://www.typescriptlang.org/docs/handbook/declaration-merging.html

PR #1549, opened 2019-09-09
fix: add guard to if & unless helpers

### LINKED ISSUE #1548: If helper can generate several syntax errors that aren't very useful
If you write a malformed `{{#if}}` helper with too many arguments, you get weird syntax errors that are different depending on just how you malformed the helper:

`{{#if anything_goes_here "string"}}`

>cannot read property 'includeZero' of undefined

`{{#if anything_goes_here undefined}}`

> Cannot read property 'hash' of undefined

`{{#if anything_goes_here null}}`

> Cannot read property 'hash' of null

Here's a JS Fiddle demonstrating these errors
https://jsfiddle.net/m5a9q3b0/3/

Obviously, the if statement examples are incorrect, but the varied weird errors that get generated are the problem. In Ghost themes we regularly see all 3 of these errors, and it's clear noone understands what they mean or how to fix them.

We just added an ugly workaround to Ghost you can [see here](https://github.com/TryGhost/Ghost/commit/708927335be96d850e4a97e4d883e0848031e681#diff-4a0ce8c3ce27b90a1098e0509a2a6a2cR42)

ErisDS:
This is a small proposed change that adds an error guard. I'm not too sure on wording though, which is basically 90% of the change 😂 

Happy to update wording and add tests etc if it's seen as an OK change 👍 

fixes #1548


nknapp:
Are you proposing this because of security relevant issues? In that case send me an email please. 
Or is this for the convenience of template authors, so that they can see errors early on?

I am OK with such a change in principle. But I think, we should add a feature flag to switch it on, because it is a breaking change (for people who made the mistake to pass more parameters to an `{{#if}}`). 

Maybe we should establish something like `Handlebars.features.checkBuiltinHelperParameters(true)` for featured flags.

I would also like to know if @wycats agrees to this, because he always wanted to have a spec before adding new features and I'm not sure if this would be a language-feature in his view.

# In case we want such a feature:

* I would say that, the check should be for `!= 2` rather than `> 2`.
* It should be added to all built-in helpers.
* If it is a feature-flag, we also need typings for the flag activation method.

nknapp:
@ErisDS despite my things to consider: If you add tests and fix the build, I will merge it.

ErisDS:
Hey @nknapp, have updated this, adding tests and taking your feedback as far as I was able to for now.

What I did first was add a bunch of cases showing these inconsistent error messages for the if, unless and with helpers & commit that. I assume you'll squash this - it's just there to demonstrate the issues.

I've switched the logic from `>2` to `!= 2` as you suggested, as I was able to find similar error cases for no arguments. 

I've then added a commit that implements the guards and shows the error messages change to become consistent and hopefully a bit clearer.

No other tests have to change for this to pass, so if there are any edge cases they are undocumented and untested.

I've not gone as far as a custom Exception class & typings - this is a bit beyond what I'm confident with just now - I'm not familiar with TypeScript. However, I do have other work I'd like to do with error handling, so time-permitting I'll revisit this whole area with some more improvements soooooon.

For now, I think this is a small change which offers some improvement 🤞

nknapp:
Hi @ErisDS. Thanks for the detailed explanation. I will have a look when I have some time.

lib/handlebars/helpers/if.js
```diff
@@ -1,7 +1,9 @@
-import {isEmpty, isFunction} from '../utils';
+import { isEmpty, isFunction } from '../utils';
+import Exception from '../exception';
 
 export default function(instance) {
   instance.registerHelper('if', function(conditional, options) {
+    if (arguments.length != 2) { throw new Exception('#if requires exactly one argument');}
```
nknapp: This really look weird. (`length != 2` leads to a message with `requires exactly one...`), but it is correct.

nknapp:
Thank you @ErisDS 

I would release this in the next minor update, if that's ok. I haven't got a real plan yet about when that will be, but I don't want to release too often anymore unless it is a critical bug.

ErisDS:
Yeah I’m in absolutely no rush 😬

PR #1551, opened 2019-09-16
wip: disallow lambdas and functions in input data...

nknapp:
...unless explicitly setting "allowUnsafeFunctionCalls"
in the compile options

* It is *NOT* backwards-compatible.
* The old, unsafe, behavior can be configured by setting "allowUnsafeFunctionCalls" to "true" in the compiler options.

@ErisDS @wycats I would like your oppinion on this. If you agree, I will add (and fix) tests and docs. Then I'll release this.

Despite the breaking change, I would only bump the patch version, as @ErisDS suggested lately.

### DIFF
--- a/components/bower.json
+++ b/components/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "handlebars",
-  "version": "4.2.0",
+  "version": "4.2.1",
   "main": "handlebars.js",
   "license": "MIT",
   "dependencies": {}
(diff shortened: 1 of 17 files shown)


nknapp:
I'm not sure if sandboxing the whole template on execution might be an alternative as well.
This article describes a method that seems to achieve that:https://blog.risingstack.com/writing-a-javascript-framework-sandboxed-code-evaluation/. It needs proxies though. I don't think it will work in IE11

nknapp:
I'm not sure if this works out. I'll close this and try again later.

nknapp:
Could you review #1559 instead

Issue #1553, opened 2019-09-18
Missing helpers after latest update


paulfalgout:
Before filing issues, please check the following points first:

- [x] Please don't open issues for security issues. Instead, file a report at https://www.npmjs.com/advisories/report?package=handlebars
- [x] Have a look at https://github.com/wycats/handlebars.js/blob/master/CONTRIBUTING.md
- [x] Read the FAQ at https://github.com/wycats/handlebars.js/blob/master/FAQ.md
- [x] Use the jsfiddle-template at https://jsfiddle.net/9D88g/47/ to reproduce problems or bugs

This will probably help you to get a solution faster. 
For bugs, it would be great to have a PR with a failing test-case.

I can try and find a way to reproduce this, but to get it started in case it's somehow obvious. In `4.1.2` everything with the below helpers worked fine.  In `4.2.0` I get `"Missing helper: "far""`...  _but_ it only seems to error when I'm instrumenting the code with istanbul.

However there are no other changes `npm i handlebars@4.1.2` fixes the issue and `@4.2.0` causes it again.

```js
import _ from 'underscore';
import Handlebars from 'handlebars/runtime';
import { findIconDefinition, icon, config } from '@fortawesome/fontawesome-svg-core';
import '@fortawesome/fontawesome-svg-core/styles.css';

config.autoAddCss = false;
config.replacementClass = 'svg-inline--fa icon';

function faHelper(prefix, iconName, { hash = {} }) {
  const faDefinition = findIconDefinition({ prefix, iconName });
  const faIcon = icon(faDefinition, hash);

  if (!faIcon || !faIcon.html) {
    throw new Error(`${ prefix }:${ iconName } fontawesome icon not loaded`);
  }
  return new Handlebars.SafeString(faIcon.html);
}

// {{far "acorn"}} -> <svg ...>
Handlebars.registerHelper({
  far: _.partial(faHelper, 'far'),
  fas: _.partial(faHelper, 'fas'),
  fal: _.partial(faHelper, 'fal'),
});

```

nknapp:
There are a couple of things that I would like you to try:

* Go to your `node_modules/handlebars` and remove the  `browser` property from the `package.json`, see if it works then.
* Replace the `browser` tag by 

```
  ...
  "browser": {
    ".": "./dist/cjs/handlebars.js",
    "./runtime": "./dist/cjs/handlebars.runtime.js"
  },
 ....
```
and see if it works then. If it doesn't then...

[...]

PR #1540, opened 2019-08-22
added browser to package.json, resolves #1102

ouijan:
Before creating a pull-request, please check https://github.com/wycats/handlebars.js/blob/master/CONTRIBUTING.md first.

Generally we like to see pull requests that

- [ ] Please don't start pull requests for security issues. Instead, file a report at https://www.npmjs.com/advisories/report?package=handlebars
- [ ] Maintain the existing code style
- [ ] Are focused on a single change (i.e. avoid large refactoring or style adjustments in untouched code if not the primary goal of the pull request)
- [ ] Have good commit messages
- [ ] Have tests
- [ ] Have the [typings](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html) (lib/handlebars.d.ts) updated on every API change. If you need help, updating those, please mention that in the PR description.
- [ ] Don't significantly decrease the current code coverage (see coverage/lcov-report/index.html)
- [ ] Currently, the `4.x`-branch contains the latest version. Please target that branch in the PR.

PR #1555, opened 2019-09-19
fix: use cjs for browser fields

paulfalgout:
We should not be using pre-built files for the `browser`.  Resolves https://github.com/wycats/handlebars.js/issues/1553

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,8 +64,8 @@
   "main": "lib/index.js",
   "types": "types/index.d.ts",
   "browser": {
-    ".": "./dist/handlebars.min.js",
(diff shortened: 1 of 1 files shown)


paulfalgout:
Related to https://github.com/wycats/handlebars.js/pull/1540
@ouijan I _think_ this PR is a nonbreaking solution with similar, perhaps improved, results for the issue you were seeing with imports.  While I think the packaged `min.js` file may function in some cases, it's breaking in others... But I _think_ these cjs cover your case and the general. If you don't mind 👀 on this.

PR #1559, opened 2019-09-21
Disallow calling "helperMissing" and "blockHelperMissing" directly

### LINKED ISSUE #1558: Disallow calling "helperMissing" and "blockHelperMissing" directly
The recent remote-code-execution exploits where misusing the helper `blockHelperMissing` in order to call methods from object prototypes that actually should not have been called.

The helpers "helperMissing" and "blockHelperMissing" are not meant to be called directly as in 

```hbs
{{helperMissing}}
{{#helperMissing}}{{/helperMissing}}
{{blockHelperMissing arg}}
{{#blockHelperMissing arg}}{{/blockHelperMissing}}
```

We can offer a flag that allows execution in case someone really wants it, but the default will be not to allow this.

Note that although calling `{{helperMissing}}` from the template is not part of the documented API and I assume it makes no sense other then to build an exploit. From this point of  view this is **not** a breaking change an will result in a minor version update only (because of the feature flag).

Even though it is not documented, I would rather try to maintain compatibility here, but since this a security-relevant change I will not do that in this case.

nknapp:
closes #1558

### DIFF
--- a/lib/handlebars/base.js
+++ b/lib/handlebars/base.js
@@ -5,7 +5,7 @@ import {registerDefaultDecorators} from './decorators';
 import logger from './logger';
 
 export const VERSION = '4.2.1';
-export const COMPILER_REVISION = 7;
+export const COMPILER_REVISION = 8;
 
 export const REVISION_CHANGES = {
   1: '<= 1.0.rc.2', // 1.0.rc.2 is actually rev2 but doesn't report it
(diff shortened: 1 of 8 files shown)


nknapp:
This PR essentially moves the `helperMissing` and `blockHelperMissing` to a `hooks`-object in the container. The compiler uses the the `hooks` object to call them when a helper is missing.

That way, the helpers cannot be called directly anymore.

nknapp:
I'll release this change now. Because I want to talk about the general issue at a user-group and I feel better if the current security issue is not working like this anymore.

PR #1570, opened 2019-10-06
Added missing type fields for some interfaces

Mindar:
While using the compiler API I ran into some issues with the typings. I wanted to use [discriminated unions](https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions) but wasn't able to do so, because the typings are missing discriminant fields. I added those for some interfaces, but didn't want to mess with some others. It might be a good idea to add those, but I'd like to have some feedback first.

It might also be a good idea to rename the interface ```Statement``` to ```StatementBase``` and then add a type called ```Statement``` which is the union of all the Statement interfaces. I don't think that change would break anything. Short example:

```typescript
interface StatementBase implements Node {}

type Statement = ContentStatement | BlockStatement | MustacheStatement | ...;

interface ContentStatement implements StatementBase {
    type: 'ContentStatement';
    value: string;
    original: StripFlags;
}

interface BlockStatement implements StatementBase {
    ...
}

...
```
That would allow for code like this:

```javascript

const program = handlebars.parse(mytemplate);

for(const statement of program) {
    switch(statement.type) {
        case 'ContentStatement':
            console.log(statement.value); // no typescript error
            break;
        ...
        default:
            break;
    }
}

```
Currently - even with this change - this is a bit more cumbersome.

Since this is my first pull request I'd love to get some feedback :)

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -270,6 +270,7 @@ declare namespace hbs {
       interface Statement extends Node {}
 
       interface MustacheStatement extends Statement {
+          type: 'MustacheStatement';
           path: PathExpression | Literal;
           params: Expression[];
           hash: Hash;
@@ -280,6 +281,7 @@ declare namespace hbs {
       interface Decorator extends MustacheStatement { }
 
       interface BlockStatement extends Statement {
+          type: 'BlockStatement';
           path: PathExpression;
           params: Expression[];
           hash: Hash;
@@ -293,6 +295,7 @@ declare namespace hbs {
       interface DecoratorBlock extends BlockStatement { }
 
       interface PartialStatement extends Statement {
+          type: 'PartialStatement';
           name: PathExpression | SubExpression;
           params: Expression[];
           hash: Hash;
@@ -301,6 +304,7 @@ declare namespace hbs {
       }
 
       interface PartialBlockStatement extends Statement {
+          type: 'PartialBlockStatement';
           name: PathExpression | SubExpression;
           params: Expression[];
           hash: Hash;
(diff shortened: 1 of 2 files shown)


nknapp:
The recommended and documented way to traverse an AST is to create a class extending the Visitor-class. I've looked at the code and believe that the typings support this properly.

The description is in docs/compiler-api.md in this repo. The example still uses ES5 syntax, but it should work using 

    class MyVisitor extends Visitor 

As well

nknapp:
I still have to think about adding these fields. It should not be a breaking change. But since people tend to use undocumented features or might still break somebody's build.

But mainly, the change encourages circumventing the official compiler API, which in the end means that it is harder to make changes to the library. It is not a good thing to have multiple ways to achieve the same thing, because all ways will have to be supported in the future.

I would like you to try extending the Visitor and report any problem that you have using this method.

Mindar:
If I use a Visitor I don't have the issue with the Statements anymore, but I still have the same issue when I try to access `someMustacheStatement.path`, which is either a `PathExpression` or `Literal` which could be one of `StringLiteral`, `NumberLiteral`, `BooleanLiteral`, `UndefinedLiteral` or `NullLiteral`.

```javascript
if(someMustacheStatement.path.type !== "PathExpression") {
    console.log(someMustacheStatement.path.original); // Typescript error
}
```

I don't think the change encourages circumventing the official compiler API. I think it documents the existing behavior. After all if an object is a `MustacheStatement` but doesn't have the string "MustacheStatement" as it's `type` there is an error. Currently this type of error is not detectable, but with my change it is.

nknapp:
Ok, you are right. It is in the docs. I hadn't seen that.

I will accept the PR, but you have to target the 4.x branch in order to get it into the current release.

Also, please append a testcase to "types/tests.ts" to make sure it stays in the typings.

PR #1571, opened 2019-10-07
added missing type fields in typings and tests for them

Mindar:
Related discussion and PR at #1570 
I wasn't able to retarget the other pull request to `4.x`, so I'm creating this new pull request. It contains some more missing type fields and a test for all of them. I'm not sure if that's the smartest way to test these fields though, maybe someone smarter than me can enlighten me?

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -270,6 +270,7 @@ declare namespace hbs {
       interface Statement extends Node {}
 
       interface MustacheStatement extends Statement {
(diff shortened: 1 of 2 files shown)


nknapp:
I think we should separate the typescript tests into separate files some time. This one large file `tests.ts` is becoming unmaintainable. I have to test somethings out before I release this, so I can't promise it will be tonight, but I will.

Issue #1576, opened 2019-10-17
Exception fields are missing in the typings


nknapp:
The file `lib/handlebars/exception.js` defines a couple of properties for exceptions.

There is also a `function Exception(...)` in `types/index.d.ts`, but it does not contain any properties.

The goal of this issue is be to define the appropriate properties on the `Exception` prototype in the typings.

kabirbaidhya:
@nknapp I am working on this one.

nknapp:
This is fixed

PR #1583, opened 2019-10-24
Add missing types for the Exception class properties

kabirbaidhya:
Resolves https://github.com/wycats/handlebars.js/issues/1576

Add missing type declarations for the properties of [`Exception`](https://github.com/wycats/handlebars.js/blob/4.x/lib/handlebars/exception.js) class. 

**What has changed?**
 - Converted `function` to a `class` so as to declare fields.
 - Declare all the properties (with the types inferred) that have been attached to the `this` object in the source [file](https://github.com/wycats/handlebars.js/blob/4.x/lib/handlebars/exception.js#L4-L58). 
 - Declare a missing second argument in the typing - `node` of type `hbs.AST.Node`.
 - Add tests for the exception typing. 

---

- [x] Have tests

---
@nknapp Not sure if this change requires a test too. If yes, please suggest and show me some examples and I can do it. Thanks!

PS: I tried `npm checkTypes` and `npm test` which all passed for me - not sure what else I need to do.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -86,15 +86,8 @@ declare namespace Handlebars {
 
   export function noConflict(): typeof Handlebars;
 
-  export interface ExceptionNode {
-      loc?: {
-        start: { line: any; column: any; };
-        end: { line: any; column: any; };
-      }
-  }
-
   export class Exception {
-      constructor(message: string, node?: ExceptionNode);
+      constructor(message: string, node?: hbs.AST.Node);
       description: string;
       fileName: string;
       lineNumber?: any;
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -97,8 +97,8 @@ declare namespace Handlebars {
       constructor(message: string, node?: ExceptionNode);
       description: string;
       fileName: string;
-      lineNumber?: string;
-      endLineNumber?: string;
+      lineNumber?: any;
+      endLineNumber?: any;
       message: string;
       name: string;
       number: number;
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -86,8 +86,15 @@ declare namespace Handlebars {
 
   export function noConflict(): typeof Handlebars;
 
+  export interface ExceptionNode {
+      loc?: {
+        start: { line: any; column: any; };
+        end: { line: any; column: any; };
+      }
+  }
+
   export class Exception {
-      constructor(message: string);
+      constructor(message: string, node?: ExceptionNode);
       description: string;
       fileName: string;
       lineNumber?: string;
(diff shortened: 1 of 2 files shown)


nknapp:
Test cases for the typings for into `types/test.ts` file.

kabirbaidhya:
OK. I'll add some.

kabirbaidhya:
@nknapp I've added some code using the defined types in the `types/test.ts` file. Please review and let me know if it makes sense. Thanks!

types/index.d.ts
```diff
@@ -86,6 +85,27 @@ declare namespace Handlebars {
...
 
+  export interface ExceptionNode {
+      loc?: {
+        start: { line: any; column: any; };
+        end: { line: any; column: any; };
+      }
+  }
+
+  export class Exception {
+      constructor(message: string, node?: ExceptionNode);
```
nknapp: The second parameter should be of type "AST.Node"

types/index.d.ts
```diff
@@ -86,6 +85,27 @@ declare namespace Handlebars {
 
   export function noConflict(): typeof Handlebars;
 
+  export interface ExceptionNode {
```
nknapp: I don't think we need this interface. The second parameter should be a `AST.Node`

types/test.ts
```diff
@@ -192,4 +192,43 @@ switch(allthings.type) {
     break;
   default:
     break;
-}
\ No newline at end of file
+}
+
+// Test exception constructor with just message.
```
nknapp: I know that it is not done this way anywhere yet, but I would wrap this test into a function. The important thing is not that it is called. The test is just used to by the TypeScript-validator.

It is not executed.

Just make sure that the main use-cases (creation with message and node, access to fields) are covered.
The tests are currently more verbose than they need to be. It is not important that `Just a message` is really this string. Assigments like the following are enough:

```typescript
function testExceptionTypings() {
   let exception: Handlebars.Exception = new Handlebars.Exception("message", {
      type: 'MustacheStatement',
      ...
   })
   let message: string = exception.message
   let lineNumber: number = exception.lineNumber
   ...
}
```

types/test.ts
```diff
@@ -192,4 +192,43 @@ switch(allthings.type) {
...
+try {
+  const exp = new Handlebars.Exception('Just a test message.');
```
nknapp: Please use longer descriptive names, no shorthands (`exception` rather than `exp`). This is also not yet done everywhere, but I will try to have an eye on such things from now on.

kabirbaidhya:
@nknapp Thanks for the detail review. I'll make the changes push again.

kabirbaidhya:
Hi @nknapp I've pushed new changes as per your suggestions. Could you please review again and let me know if this looks good? Thanks!

nknapp:
Superb. Merging now.

PR #1584, opened 2019-10-25
Add ability to disable whitespace control: `Handlebars.parseWithoutProcessing`

rwjblue:
When authoring tooling that parses Handlebars files and emits Handlebars files, you often want to preserve the **exact** formatting of the input. The changes in this commit add a new method to the `Handlebars` namespace: `parseWithoutProcessing`. Unlike, `Handlebars.parse` (which will mutate the parsed AST to apply whitespace control) this method will parse the template and return it directly (**without** processing 😉).

For example, parsing the following template ([feel free to review the AST on ASTExplorer](https://astexplorer.net/#/gist/f850c5457b808756107c0bc5ad226989/b0366df64be5c01b5f9eadd93e795b3b9d65d2c7)):

```hbs
 {{#foo}}
   {{~bar~}} {{baz~}}
 {{/foo}}
```

Using `Handlebars.parse`, the AST returned would have truncated the following whitespace:

* The whitespace prior to the `{{#foo}}`
* The newline following `{{#foo}}`
* The leading whitespace before `{{~bar~}}`
* The whitespace between `{{~bar~}}` and `{{baz~}}`
* The newline after `{{baz~}}`
* The whitespace prior to the `{{/foo}}`

When `Handlebars.parse` is used from  `Handlebars.precompile` or `Handlebars.compile`, this whitespace stripping is **very** important (these behaviors are intentional, and generally lead to better rendered output).

When the same template is parsed with `Handlebars.parseWithoutProcessing` none of those modifications to the AST are made. This enables "codemod tooling" (e.g. `prettier` and `ember-template-recast`) to preserve the **exact** initial formatting.

Prior to these changes, those tools would have to _manually_ reconstruct the whitespace that is lost prior to emitting source.

### DIFF
--- a/lib/handlebars/compiler/base.js
+++ b/lib/handlebars/compiler/base.js
@@ -8,7 +8,7 @@ export { parser };
 let yy = {};
 extend(yy, Helpers);
 
-export function parse(input, options) {
+export function parseWithoutProcessing(input, options) {
   // Just return if an already-compiled AST was passed in.
   if (input.type === 'Program') { return input; }
 
@@ -19,6 +19,14 @@ export function parse(input, options) {
     return new yy.SourceLocation(options && options.srcName, locInfo);
   };
 
+  let ast = parser.parse(input);
+
+  return ast;
+}
+
+export function parse(input, options) {
+  let ast = parseWithoutProcessing(input, options);
   let strip = new WhitespaceControl(options);
-  return strip.accept(parser.parse(input));
+
+  return strip.accept(ast);
 }
--- a/spec/ast.js
+++ b/spec/ast.js
@@ -123,12 +123,94 @@ describe('ast', function() {
      });
   });
 
+  describe('whitespace control', function() {
+    describe('parse', function() {
+      it('mustache', function() {
+        let ast = Handlebars.parse('  {{~comment~}} ');
+
+        equals(ast.body[0].value, '');
+        equals(ast.body[2].value, '');
+      });
+
+      it('block statements', function() {
+        var ast = Handlebars.parse(' {{# comment~}} \nfoo\n {{~/comment}}');
+
+        equals(ast.body[0].value, '');
+        equals(ast.body[1].program.body[0].value, 'foo');
+      });
+    });
+
+    describe('parseWithoutProcessing', function() {
+      it('mustache', function() {
+        let ast = Handlebars.parseWithoutProcessing('  {{~comment~}} ');
+
+        equals(ast.body[0].value, '  ');
+        equals(ast.body[2].value, ' ');
+      });
+
+      it('block statements', function() {
+        var ast = Handlebars.parseWithoutProcessing(' {{# comment~}} \nfoo\n {{~/comment}}');
+
+        equals(ast.body[0].value, ' ');
+        equals(ast.body[1].program.body[0].value, ' \nfoo\n ');
+      });
+    });
+  });
+
   describe('standalone flags', function() {
     describe('mustache', function() {
       it('does not mark mustaches as standalone', function() {
         var ast = Handlebars.parse('  {{comment}} ');
         equals(!!ast.body[0].value, true);
         equals(!!ast.body[2].value, true);
+      });
+    });
+    describe('blocks - parseWithoutProcessing', function() {
+      it('block mustaches', function() {
+        var ast = Handlebars.parseWithoutProcessing(' {{# comment}} \nfoo\n {{else}} \n  bar \n  {{/comment}} '),
+          block = ast.body[1];
+
+        equals(ast.body[0].value, ' ');
+
+        equals(block.program.body[0].value, ' \nfoo\n ');
+        equals(block.inverse.body[0].value, ' \n  bar \n  ');
+
+        equals(ast.body[2].value, ' ');
+      });
+      it('initial block mustaches', function() {
+        var ast = Handlebars.parseWithoutProcessing('{{# comment}} \nfoo\n {{/comment}}'),
+          block = ast.body[0];
+
+        equals(block.program.body[0].value, ' \nfoo\n ');
+      });
+      it('mustaches with children', function() {
+        var ast = Handlebars.parseWithoutProcessing('{{# comment}} \n{{foo}}\n {{/comment}}'),
+          block = ast.body[0];
+
+        equals(block.program.body[0].value, ' \n');
+        equals(block.program.body[1].path.original, 'foo');
+        equals(block.program.body[2].value, '\n ');
+      });
+      it('nested block mustaches', function() {
+        var ast = Handlebars.parseWithoutProcessing('{{#foo}} \n{{# comment}} \nfoo\n {{else}} \n  bar \n  {{/comment}} \n{{/foo}}'),
+          body = ast.body[0].program.body,
+          block = body[1];
+
+        equals(body[0].value, ' \n');
+
+        equals(block.program.body[0].value, ' \nfoo\n ');
+        equals(block.inverse.body[0].value, ' \n  bar \n  ');
+      });
+      it('column 0 block mustaches', function() {
+        var ast = Handlebars.parseWithoutProcessing('test\n{{# comment}} \nfoo\n {{else}} \n  bar \n  {{/comment}} '),
+            block = ast.body[1];
+
+        equals(ast.body[0].omit, undefined);
+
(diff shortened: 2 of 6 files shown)


scalvert:
This will definitely address a pain point WRT recast and whitespace, that I myself have definitely encountered. Thanks for this. 🙏

nknapp:
I can see the need, but I don't think that a compile-option is the right way to go here. A compile-option would suggest that the use-case is compiling and executing a template with whitespace-control disabled.

But all you want is to build a clean, unmodified AST, correct? 

I think a second "parse" function should be exposed for the purpose of building tools. Maybe there are other modifications that are useful for tooling, that could be built into the second function.

Consequently, the test-cases should not be along the lines of "shouldCompileTo" but more like the tests in the "ast.js"-spec: Build an AST and compare it to the expected one.

Does that make sense?

rwjblue:
> I don't think that a compile-option is the right way to go here

Hmm, this is a `ParseOption`, not a compile option, right? Am I thinking about the structure incorrectly?

> I think a second "parse" function should be exposed for the purpose of building tools. Maybe there are other modifications that are useful for tooling, that could be built into the second function.

Sure, that seems totally fine to me as well. What do you think it should be called? Where should it go?

> Consequently, the test-cases should not be along the lines of "shouldCompileTo" but more like the tests in the "ast.js"-spec: Build an AST and compare it to the expected one.

Sounds good, thank you for reviewing!

rwjblue:
@nknapp - I just pushed another commit that separates `parse` into two functions (one that parses the input template string, and the other that leverages the raw parse output and adds whitespace control), as well as adding a number of additional tests to `spec/ast.js`.

nknapp:
How about `parsePlainAST`, `parseRawAST`, `parseUnprocessedAST`

But "raw" is already taken for '{{{{raw-blocks}}}} {{{{/raw-blocks}}}}`. This might be confusing.
I am not sure about this either. The name might be only temporary, because we have plans to extract the parser into a separate package. When we do that, we can choose proper names for the function or even leave the white-space processing out of the package completely.

But we don't have a timeline. I can't make one with the small amount of work I can put into this every day.

lib/handlebars/compiler/base.js
```diff
@@ -8,7 +8,7 @@ export { parser };
 let yy = {};
 extend(yy, Helpers);
 
-export function parse(input, options) {
+export function parseAST(input, options) {
```
nknapp: I think we should call it "parseWithoutProcessing".

types/index.d.ts
```diff
@@ -47,8 +47,9 @@ declare namespace Handlebars {
   }
 
   export interface ParseOptions {
-      srcName?: string,
-      ignoreStandalone?: boolean
+      srcName?: string;
+      ignoreStandalone?: boolean;
+      disableWhitespaceControl?: boolean;
```
nknapp: Typings for the change are still missing here.

A test for the typings should go into `types/test.ts`. Please wrap the test into a new function like 

```ts
function testParseWithoutProcessing() {
   .... // call new function and make sure the the typings match
}
```

nknapp: There should also be mention of the function in the `docs/compiler-api.md` file.

lib/handlebars/compiler/base.js
rwjblue: Sounds good! Updated.

types/index.d.ts
rwjblue: Awesome, thank you for the pointer on how to test the types. I've added a function that does this, mind reviewing to confirm I did it correctly?

rwjblue:
Thanks for your time @nknapp! I've updated with your latest round of feedback:

* Rename method to `parseWithoutProcessing`
* Update types for new method
* Add `types/test.ts` for new API

spec/ast.js
```diff
@@ -131,6 +131,54 @@ describe('ast', function() {
...
+            block = ast.body[1];
+
+        equals(ast.body[0].omit, undefined);
+
+        equals(block.program.body[0].value, ' \nfoo\n ');
+        equals(block.inverse.body[0].value, ' \n  bar \n  ');
+
+        equals(ast.body[2].value, ' ');
+      });
+    });
```
nknapp: I haven't noticed this before, but have you added test that specifically tests that whitespace-control is NOT executed?

rwjblue: The tests here confirm that some of the things that `lib/handlebars/compiler/whitespace-control.js` does are *not* done, but you are correct that I don't have a test using ` {{~foo~}}  ` and the like. Will add something like that shortly.

rwjblue:
OK, just pushed another couple of changes:

* Added documentation for `parseWithoutProcessing` to `docs/compiler-api.md`
* Add explicit tests to `spec/ast.js` for using whitespace control (with `parse` and `parseWithoutProcessing`)

spec/ast.js
rwjblue: Just added a few more tests to this file that use whitespace control (`~`) directly. Confirming that `parse` and `parseWithoutProcessing` have the appropriate behaviors here.

nknapp:
Thank you. I will have another glance but all in all the PR looks good to me now.

types/test.ts
```diff
@@ -192,4 +192,13 @@ switch(allthings.type) {
     break;
   default:
     break;
-}
\ No newline at end of file
+}
+
+function testParseWithoutProcessing() {
+  // call new function and make sure the the typings match
+  const parsedTmpl = Handlebars.parseWithoutProcessing('<p>Hello, my name is {{name}}.</p>', {
```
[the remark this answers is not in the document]
rwjblue: Gotcha, will update (and squash commits) shortly.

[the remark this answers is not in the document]
rwjblue: Updated!

rwjblue:
Just pushed another round of updates:

* Fixed issues in `types/test.ts`
* Squashed commits down to one
* Updated description and commit messages to match current implementation

@nknapp - Please let me know if there is anything else that needs some tweaks. 😸

nknapp:
I would have done the squashing myself, but well OK. 

I noticed that unsquashed commits are much easier to review.

rwjblue:
> I haven't decided on the date of the next release yet. Is this urgent for you or can it wait another couple of weeks.

There are some bugs that are a bit difficult to work around in the codemod tooling that I am maintaining (things like https://github.com/ember-template-lint/ember-template-recast/issues/155) that this helps resolve (once updated in `@glimmer/syntax`). Now that I know how the future versions of `handlebars` will support this, I can probably do a (slightly hacky) work around while we wait for 4.5.0. 

tldr; I _think_ I can "polyfill" this new method in my codemod tooling, if I run into issues there though I will let you know...

rwjblue:
> I would have done the squashing myself, but well OK.

Gotcha. I also wanted to fix up the commit message a bit (the main commit message was still talking about the original option I proposed).

> I noticed that unsquashed commits are much easier to review.

Yes, I whole heartedly agree. The GitHub UI makes reviewing the delta between commits super simple, but when folks continually squash / amend that workflow is broken. FWIW, I have had _some_ luck with the new "mark file as viewed" functionality in the GitHub pull request UI (this works across squashes), but it's still not as nice as leaving it unsquashed.

nknapp:
Before you start building workarounds... I would like wait for #1583 to be ginishrd. Then I could probably release 4.5 on the next weekend.

It won't be more than a couple of weeks at most.  Maybe you can use a github dependency until then.

nknapp:
If you want this for ember, you probably need a backport to 3.x, don't you?

nknapp:
I might be able to release this evening. Do you know if you need the featured in 4.x or 3.x?

rwjblue:
> If you want this for ember, you probably need a backport to 3.x, don't you?

We are using v4 over in `@glimmer/syntax` (which is what Ember uses for parsing).

> It won't be more than a couple of weeks at most.

Ya, definitely no worries on timing from me. Thank you again for all of your help getting this into a good state.

nknapp:
> We are using v4 over in @glimmer/syntax (which is what Ember uses for parsing).

Really? A while ago @wycats asked for my opinion on how to proceed with the parser. He said Ember was using an older version and he was thinking about how to converge the code-base again, because Handlebars has features that are not used by Ember and Ember has other features like components.

But OK. It's a good to know that you are using the current version.

nknapp:
I have release 4.5 now. It was just good to have all this wrapped up. (will take a break now).

Issue #1574, opened 2019-10-09
Deprecate decorators


nknapp:
The decorator API currently exposes a good portion of internal Handlebars methods (like the  `container`-object) that should not be part of the official API. This makes it difficult to write a formal specification and port this API to implementations in other programming languages.

In the next minor version of Handlebars, the API will be deprecated and warning messages will be written to the console, when attempting to register a decorator.

**If you use decorators, please add a comment here** so we can discuss alternatives.

allouis:
Out of interest, would only the registering of custom decorators be deprecated, or would support for all of them eventually be deprecated/removed?

I'm specifically interested in the usage of the `inline` decorator inside of partial blocks - like the example from the current docs:
```hbs
{{#> layout}}
  {{#*inline "nav"}}
    My Nav
  {{/inline}}
  {{#*inline "content"}}
    My Content
  {{/inline}}
{{/layout}}
```

nknapp:
Only registering custom decorators will be deprecated. I have discussed this and other matters with Yehuda and we won't even remove this in 5.0, more like 6.0

And I don't think 5.0 is going to come this year.

But I still need to know if anybody object, because if somebody does, I would like to fond a solution. We may offer a similar API. My only problem with the current implementation is the amount of exposure of private code.

cthorner:
Is there an alternative to decorators? Or just use helpers more extensively? Seems like this is going to add some complexity to templates already using decorators.

nknapp:
@cthorner Thank you for joining the discussing. If you need decorators, what parts of the API you are using?

I am bothered mostly about the `container` being exposed. If you are using any function from the container I would like to know which ones and for which purposes. In any case, I would be interested in example decorators, because all that I have seen so far is the `inline`-decorator and a minimal textbook-example from [sitepoint](https://www.sitepoint.com/using-inline-partials-and-decorators-with-handlebars-4-0/) as well as in the test-cases. Apart from `inline` I have no idea what kind of decorators are used in production anywhere. If you can't publish your code, you can also send me an email.

We can develop alternatives. But I need to know the use-cases.

jbboehr:
(slightly off-topic) I haven't implemented decorators in my port/implementation ([handlebars.c](https://github.com/jbboehr/handlebars.c)) in part due to this, mostly due to it being really hairy in C, but also:

[...]

Issue #1561, opened 2019-09-24
4.3.0 compiler change breaks Semver.


oculus42:
The compiler change with 4.3.0 breaks semver so precompiled templates in internal dependencies are unable to execute.

Please make compiler changes major version updates.

oculus42:
In the short term we are pinning 4.2.1 in place of recompiling all packages and contacting customers that produce Handlebars content.

nknapp:
Hi @oculus42,

I am aware that this was a breaking change and I have mentioned that in the release notes. However, this change was necessary to eliminate a new variant of the Remote Code Execution exploit that appear in the beginning of this year. 

It was hard to fix the leak without making any breaking changes, but if I had published the change as 5.0, nobody would be aware of the issue.

I will have a look if there is a way to fix the exploit without requiring to increase the compiler version.
But if that does not work, I would like urge you to see if you cannot compile your templates with a more recent precompiler.

oculus42:
Hey @nknapp,

Thanks for the reply. I will likewise take a look this evening and see if I can help in the effort to produce a non-breaking change.

nomoon:
I work on a big, unwieldy, institutional SDK that I can't control all of the parts of (which are at worst semver scoped to the latest minor) and this change has cost me hours trying to find a workaround to.

nknapp:
I am sorry for everyone who had extra work because of this release. I know that it is frustrating to search for a bug for hours, and to find out that some library has made a breaking change.

My problem is that I don't have a communication channel to reach all Handlebars users... And I have a major and publicly known security issue to resolve and no team to review my changes. This release needed to be published. And it needed to be a 4.x because otherwise a lot of systems would remain unpatched (because everybody scopes to the latest minor).

@oculus42 I have a fix ready, but it might take until tomorrow, because the build is still running and I need to go to bed. If everything works, the runtime will accept templates with compiler version 7 and automatically switch to the fallback mode.

nknapp:
the fix on npm now.

nomoon:
@nknapp Much appreciated!

nknapp:
Stayed up longer than I wanted to though...

radiolips:
@nknapp I'd like to think I'm speaking for everyone when I say thank you. You caused a bit of a disruption for my team, too, but you tried to do the most responsible thing you could, then threw in a fix hours after @oculus42 opened this ticket. Open source stuff is a lot of work. Thanks for staying up late to get this done for us.

[...]

PR #1587, opened 2019-10-28
feat: deprecate custom decorators

nknapp:
This PR marks the "registerDecorator" is now deprecated. The
decorator API exposes a lot of internal structures that
should not be part of the official API. An alternative
API can be discussed at #1574
- "Handlebars.registerDecorator" now logs a warning
- "Handlebars.registerDecoratorWithoutDeprecationWarning"
  can be used in order to omit the warning, when
  developers have read the notice.
  (The inline-decorator is using this function now)

I would value comments about whether this is the correct way to introduce a deprecation.
My thoughts:
- Pro: People only read a changelog when they have build failures. A deprecation notice in the build 
 log might be more visible than an entry in the release-notes
- Contra: The warning will probably appear in the browser-console of end-users. This may not be
  what developers want. In some cases, developer my have a hard time removing the warnings,
 because they don't have direct access to the code that uses Handlebars. ([like this comment suggests](https://github.com/wycats/handlebars.js/issues/1561#issuecomment-534760915))
- Contra: The message may not be as visible as I intend, if it is just in the build log and in the browser-console.

### DIFF
--- a/lib/handlebars/base.js
+++ b/lib/handlebars/base.js
@@ -63,6 +63,11 @@ HandlebarsEnvironment.prototype = {
   },
 
   registerDecorator: function(name, fn) {
+    this.log('warn', 'Custom decorators are deprecated and will be removed in the future. Join the discussion at https://github.com/wycats/handlebars.js/issues/1574');
+    this.registerDecoratorWithoutDeprecationWarning(name, fn);
+  },
+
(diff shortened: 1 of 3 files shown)


nknapp:
@wycats, @ErisDS opinions?

PR #1594, opened 2019-11-05
Prevent access to other prototype properties

nknapp:
The npm security advisories [755](https://www.npmjs.com/advisories/755) and [1164](https://www.npmjs.com/advisories/1164) were both caused by access to the `constructor` property. The initial fix was circumvented in advisory (1164) by the use of `__defineGetter__` and `__defineSetter`.

In order to prevent other attacks in the future, this PR restricts access to properties in two way:

* if the compile-option `allowNonHelperFunctionCall` is set to `false`, any use of properties of the input object as function-calls with parameters is prohibited. Lambdas are still allowed.
* Access to the properties `__defineGetter__`, `__defineSetter` and `__proto__` are only allowed if they are enumerable on their parent. The list is configurable through the compile-option `propertyMustBeEnumerable`.

```javascript
Handlebars.compile('{{__defineGetter__}}', {
   propertyMustBeEnumerable: {
      "__defineGetter__": false,
      "push": true
   }
})
```

The main question is: **Which other properties should be restricted by default?**

* push, pop, fill, shift, unshift
* bind, call, apply

I don't want to cause trouble by creating unnecessary restrictions, but I also want to prevent other security leaks.


nknapp:
@skinny85 could you have a look at this?

skinny85:
@nknapp while I appreciate the vote of confidence, I think this PR is way beyond my knowledge of the subtleties of JavaScript 😱

One general comment: I think the PR description does a great job of justifying why is the change needed. I think it would be great if a similar description made its way into the comments in the code itself - I think future maintainers of this codebase (which might very well be you 😉) will appreciate it when they see this a year (or later) from now :).

Issue #1598, opened 2019-11-08
Performance degradation for partials since v4.3.0


ole-martin:
Hi there,
We got an issue where upgrading handlebars beyond v4.2.1 introduces a major performance-impact on our site. We use handlebars server-side and since v4.3.0 partials do not cache their compile-function properly it is leading to a lot of overhead in parsing & building the compile functions again and again. (Total response-time goes from about 75ms to 250ms on the average request)

### Underlying issue:
With v4.3.0+ any non-precompiled partials registered with handlebars.registerPartial do not cache their compiled value beyond the current template scope, causing them to be recompiled every time they are used for the first time within a template. 

### The cause:
I've narrowed it down and the change that is causing it is this line:
https://github.com/wycats/handlebars.js/commit/2078c727c627f25d4a149962f05c1e069beb18bc#diff-cccd9fbbd9d8ac0bff571e9c8b0ee9deL164-R159
The practical implications is that when no options.partials exists, handlebars still creates a shallow-copy of env.partials, while before container.partials would be set to reference the env.partials directly. 
I haven't been able to prove exactly why this happens, but my assumption is that as the partials are not compiled at the time of the shallow-copy, the env.partials contains strings which never gets updated since strings are not copied by reference (thus when container.partials['my-partial'] is updated to the compiled function, the original env.partials reference is not affected)

### Proposed solution:
(Note: I'm assuming removing the possibility of directly assigning env.partials to container.partials was not an intentional change to prevent the security issues in v4.2.1)
In runtime.js change L161
`container.partials = Utils.extend({}, env.partials, options.partials);`
into
```
        if (!options.partials || !env.partials) {
          container.partials = options.partials || env.partials;
        } else {
          container.partials = Utils.extend({}, env.partials, options.partials);
        }
```
It's not a perfect solution, as if there was both options.partials and env.partials defined it would still shallow-copy env.partions, but this should be more consistent with pre v4.3.0 behavior and even speed up the runtime performance as it no longer needs to loop through and shallow-copy env.partials if no options.partials are defined.

ole-martin:
While I'm unable to verify that this fiddle actually reproduces this behavior, this setup should hopefully be able to reproduce the issue if needed:
https://jsfiddle.net/mtud81p3/2/

nknapp:
I don't think I intended this change for partials. But I didn't understand why the "merge" function was doing all those extra struff, so I just removed it for the sake of cleaner code.

[...]

PR #1599, opened 2019-11-08
fix(runtime.js): partials performance fix

ole-martin:
Reintroduce assigning env.partials to container.partials if no options.partials is set. Also allow assigning options.partials to container.partials if only that one exists. This solves https://github.com/wycats/handlebars.js/issues/1598


ole-martin:
Bah, forgot to base on 4.x branch I guess, creating a new one soon

PR #1600, opened 2019-11-08
fix(runtime.js): partials compile not caching

ole-martin:
Reintroduce assigning env.partials to container.partials if no options.partials is set. Also allow assigning options.partials to container.partials if only that one exists. This solves https://github.com/wycats/handlebars.js/issues/1598

### DIFF
--- a/spec/compiler.js
+++ b/spec/compiler.js
@@ -108,14 +108,14 @@ describe('compiler', function() {
       equal(/return ""/.test(Handlebars.precompile('')), true);
     });
   });
-  
+
   describe('Compile Partials Correctly', function() {
     var compileCount = 0;
     var origTemplateFunc = Handlebars.template;
     before(function() {
-      Handlebars.template = function(...params) {
+      Handlebars.template = function(params) {
         compileCount++;
-        return origTemplateFunc(...params);
+        return origTemplateFunc(params);
       };
       Handlebars.registerPartial({
         'dude': 'I am a partial'
@@ -123,7 +123,7 @@ describe('compiler', function() {
     });
     after(function() {
       Handlebars.template = origTemplateFunc;
-      // Handlebars.unregisterPartial('dude');
+      Handlebars.unregisterPartial('dude');
     });
 
     it('should only compile global partials once', function() {
--- a/spec/compiler.js
+++ b/spec/compiler.js
@@ -108,29 +108,4 @@ describe('compiler', function() {
       equal(/return ""/.test(Handlebars.precompile('')), true);
     });
   });
-
-  describe('Compile Partials Correctly', function() {
-    var compileCount = 0;
-    var origTemplateFunc = Handlebars.template;
-    before(function() {
-      Handlebars.template = function(params) {
-        compileCount++;
-        return origTemplateFunc(params);
-      };
-      Handlebars.registerPartial({
-        'dude': 'I am a partial'
-      });
-    });
-    after(function() {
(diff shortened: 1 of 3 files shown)


andersk:
Perhaps `helpers` and `decorators` should also go back to using `merge` like they did before 2078c727c627f25d4a149962f05c1e069beb18bc?

ole-martin:
@andersk Afaik they aren't compiled, so there won't be a big difference like with partials, but it will potentially have some speed-up as it will skip the shallow-copy step if not needed. @nknapp what do you think?

nknapp:
I think the overhead of the shallow-copy step is negligible. 

#1594 is also related to that place. It makes sure that the helpers-object has a "null"-prototype, because some exploits have used "__defineGetter__" to do bad things and the "helpers"-object is accessible from the template.

Merging here makes sure that the "helpers"-option is not used "as is" and together with #1594 it ensure the null-prototype (`Object.create(null)`) for this object, so: No, I would like to leave helpers and decorators as they are.

 @ole-martin could you rename the merge-function to something that explains what the function is doing. It's not a simple "merge", but a "merge if needed" or a "first non-null or merge" or something like that. If you find a good name, short yet describing what the function does, please take it. But "firstNonNullOrMerged" would be fine to.

I was confused the first time I read that code and my intention (over time) is to make the Handlebars code more explanatory and clean (not by comments, but by good naming).

ole-martin:
@nknapp Updated function name to "mergeIfNeeded" now 👍

nknapp:
@ole-martin forgive me for delaying this further again. Although I had added some change requests, I haven't had time for a full review yet. And now it is getting too late again. But I will look...

spec/compiler.js
```diff
@@ -108,4 +108,29 @@ describe('compiler', function() {
       equal(/return ""/.test(Handlebars.precompile('')), true);
     });
   });
+
+  describe('Compile Partials Correctly', function() {
+    var compileCount = 0;
+    var origTemplateFunc = Handlebars.template;
+    before(function() {
+      Handlebars.template = function(params) {
```
nknapp: I have added chai and sinon to the dev-dependencies on the current 4.x. 

Please rebase to the lastest 4.x. you will then be able to use a [sinon-spy](https://sinonjs.org/releases/latest/spies/) to count the template invocations.

I would create the spy in the `it()`-block and call `sinon.restore()` in an  `afterEach()` block.

spec/compiler.js
```diff
@@ -108,4 +108,29 @@ describe('compiler', function() {
...
+        compileCount++;
+        return origTemplateFunc(params);
+      };
+      Handlebars.registerPartial({
+        'dude': 'I am a partial'
+      });
+    });
+    after(function() {
+      Handlebars.template = origTemplateFunc;
+      Handlebars.unregisterPartial('dude');
```
nknapp: If think it is better to use "Handlebars.create()" to create a new instance (in the `it`-block) instead.
Please try if this works.

ole-martin:
@nknapp I've moved the test to regressions.js and changed it to use sinon-spy & Handlebars.create(). I've also verified that without the changes to runtime the test fails with 
```
1) Regressions
       GH-1598: Performance degradation for partials since v4.3.0
         should only compile global partials once:
     Error: '4' should === '3'
```

nknapp:
Great, I'll have another look this evening. But I think its fine now. Thanks.

Issue #1616, opened 2019-11-25
Precompiling Templates dead link


DanielOberlechnerBrandnamic:
Hi,
I tried to click on the link in the main description but it is a dead link.
Someone should change it.

Thank you. :)

"Precompiling Templates
Handlebars allows templates to be precompiled and included as javascript code rather than the handlebars template allowing for faster startup time. Full details are located here."

nknapp:
Can you try this again? I have deployed the new documentation-page on Sunday. Maybe the old page is still cached in your browser.

DanielOberlechnerBrandnamic:
I tried it again even in different browsers but it is still there. (cleared the cache in Chrome but no changes) ... 
There must be something wrong ...

Greets

nknapp:
Could you post a screenshot of the page that contains the link?

DanielOberlechnerBrandnamic:
![error github handlebars](https://user-images.githubusercontent.com/57362803/69937825-05bf3880-14dc-11ea-9f68-5d6dcb6f906c.PNG)

Here you can see the defect link ...

Greets,
Daniel

nknapp:
Ah, I thought you where on the homepage, not on in a file on GitHub. You are right, that's indeed not fixed, but the linke PR will do it.

nknapp:
should be corrected by now

PR #1620, opened 2019-12-02
Update readme.md with updated links

ole-martin:
I noticed there was several dead links in the readme due to the new doc being released. Updated links to their new locations
Fix http->https where possible and update links doc links.
Fixes issue: https://github.com/wycats/handlebars.js/issues/1616

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -51,7 +51,7 @@ Full documentation and more examples are at [handlebarsjs.com](https://handlebar
 Precompiling Templates
 ----------------------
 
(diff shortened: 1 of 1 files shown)


ole-martin:
@nknapp Could you verify that I've updated this with the correct links?

nknapp:
Thanks for these updates. I wonder if I should have been more cautious about preserving URLs in the new site.

Maybe we can move content from the README and CONTRIBUTING to handlebarsjs.com, it might be easier to maintain that way.

PR #1633, opened 2020-01-08
feat: access control to prototype properties via whitelist

### LINKED ISSUE #1628: sub, sup, bold and small helpers undocumented?
I recently came across some weird behaviour where the template `{{#url}}{{small}}{{/url}}` when passed the data `{ url: 'https://google.com' }` wraps the url in `small` html tags, initially I thought this is just built-in helpers, but looking at the docs it is not documented as one. Furthermore, when logging the built-in helpers and partials `small` is not referenced anywhere, I also found the same to be true for `sub`, `sup` and `bold`. Searching the codebase for these terms also yields nothing. I am left very confused. Where is this behaviour coming from? and is it possible to turn it off?

I wrote a little JSFiddle to demonstrate the issue: https://jsfiddle.net/d0ve69xL/

### LINKED ISSUE #1631: Placeholder

- [ ] Please don't open issues for security issues. Instead, file a report at https://www.npmjs.com/advisories/report?package=handlebars
- [ ] Have a look at https://github.com/wycats/handlebars.js/blob/master/CONTRIBUTING.md
- [ ] Read the FAQ at https://github.com/wycats/handlebars.js/blob/master/FAQ.md
- [ ] Use the jsfiddle-template at https://jsfiddle.net/4nbwjaqz/4/ to reproduce problems or bugs

nknapp:
Disallow access to prototype properties and methods by default.
Access to properties is always checked via
`Object.prototype.hasOwnProperty.call(parent, propertyName)`.

New runtime options:
- **allowedProtoMethods**: a string-to-boolean map of property-names that are allowed if they are methods of the parent object.
- **allowedProtoProperties**: a string-to-boolean map of property-names that are allowed if they are properties but not methods of the parent object.

```js
const template = handlebars.compile('{{aString.trim}}')
const result = template({ aString: '  abc  ' })
// result is empty, because trim is defined at String prototype
```

```js
const template = handlebars.compile('{{aString.trim}}')
const result = template({ aString: '  abc  ' }, {
  allowedProtoMethods: {
    trim: true
  }
})
// result = 'abc'
```

Implementation details: The method now "container.lookupProperty"
handles the prototype-checks and the white-lists. It is used in
- JavaScriptCompiler#nameLookup
- The "[lookup](https://github.com/wycats/handlebars.js/blob/33a3b46bc205f768f8edbc67241c68591fe3472c/lib/handlebars/helpers/lookup.js#L5-L4)"-helper (passed to all helpers as "options.lookupProperty")
- The "lookup" function at the container, which is used for recursive lookups in "compat" mode

Compatibility:
- **Old precompiled templates work with new runtimes**: The "options.lookupPropery"-function is passed to the helper by a wrapper, not by the compiled templated.
- **New templates work with old runtimes**: The template contains a function that is used as fallback if the "lookupProperty"-function cannot be found at the container. However, the runtime-options "allowedProtoProperties" and "allowedProtoMethods" only work with the newest runtime.

BREAKING CHANGE:
- access to prototype properties is forbidden completely by default

closes #1631
closes #1628

### DIFF
--- a/lib/handlebars/compiler/javascript-compiler.js
+++ b/lib/handlebars/compiler/javascript-compiler.js
@@ -2,7 +2,6 @@ import { COMPILER_REVISION, REVISION_CHANGES } from '../base';
 import Exception from '../exception';
 import { isArray } from '../utils';
 import CodeGen from './code-gen';
-import { dangerousPropertyRegex } from '../helpers/lookup';
 
 function Literal(value) {
   this.value = value;
@@ -13,27 +12,8 @@ function JavaScriptCompiler() {}
 JavaScriptCompiler.prototype = {
   // PUBLIC API: You can override these methods in a subclass to provide
   // alternative compiled forms for name lookup and buffering semantics
-  nameLookup: function(parent, name /* , type*/) {
-    if (dangerousPropertyRegex.test(name)) {
-      const isEnumerable = [
-        this.aliasable('container.propertyIsEnumerable'),
-        '.call(',
-        parent,
-        ',',
-        JSON.stringify(name),
-        ')'
-      ];
-      return ['(', isEnumerable, '?', _actualLookup(), ' : undefined)'];
-    }
-    return _actualLookup();
-
-    function _actualLookup() {
-      if (JavaScriptCompiler.isValidJavaScriptVariableName(name)) {
-        return [parent, '.', name];
-      } else {
-        return [parent, '[', JSON.stringify(name), ']'];
-      }
-    }
+  nameLookup: function(parent, name /*,  type */) {
+    return this.internalNameLookup(parent, name);
   },
   depthedLookup: function(name) {
     return [this.aliasable('container.lookup'), '(depths, "', name, '")'];
@@ -69,6 +49,12 @@ JavaScriptCompiler.prototype = {
     return this.quotedString('');
   },
   // END PUBLIC API
+  internalNameLookup: function(parent, name) {
+    this.lookupPropertyFunctionIsUsed = true;
+    return ['lookupProperty(', parent, ',', JSON.stringify(name), ')'];
+  },
+
+  lookupPropertyFunctionIsUsed: false,
 
   compile: function(environment, options, context, asObject) {
     this.environment = environment;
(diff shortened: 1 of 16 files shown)


Gerrit0:
@nknapp what would be your recommendation for projects which make heavy use of templates with helper methods attached to the prototype? TypeDoc uses handlebars and passes one of many different model classes to a template that will then render it.

For now, I'm going to pin TypeDoc to an older version of Handlebars with support for prototype properties, but this really isn't a long term solution.

nknapp:
This change was the response to a reported security issue, actually. I don't want to give details until it is disclosed, but I have spent a good part of the last year, patching holes, and they kept popping up again and again. This is only sustainable solution to this problem.

Gerrit0:
An opt-out would be ideal for my use case. TypeDoc already runs arbitrary user code when rendering themes, so the mentioned attack is a non-issue. A new option or being able to set `allowedProtoMethods` to `true` would work great!

I have been meaning to attack the themes and clean them up, maybe it is time to move that issue up the priority list and remove the need for prototype access entirely. It would be a lot of work, but perhaps that time has come.

nknapp:
see #1635 

Please have a look and comment.

Implementing this was is not so much work now, because the general logic of runtime-options controlling this behaviour is already there.

nknapp:
Handlebars 4.7.0 has been release with options to disable prototype restrictions:
https://handlebarsjs.com/api-reference/runtime-options.html#options-to-control-prototype-access

tmaiaroto:
Would `JSON.parse(JSON.stringify(objWithGetter))` not have worked as a safe default behavior? When passing templates responses from ORMs (for example Sequelize ORM), you'll now need to get a plain object instead of being able to leverage its getters, eager loading, etc. In the case of Sequelize, it was a quick fix to call `toJSON()` but it sure was confusing to track this down and the minor version represented an unexpected break for something extremely common.

nknapp:
@tmaiaroto do you mean: as safe default behaviour instead of test hasOwnProperty for each access? Or instead of the changes I made in 4.7?

Gerrit0:
`JSON.parse(JSON.stringify(obj))` is far from safe. JSON doesn't handle recursive data structures, while Handlebars does support them.

tmaiaroto:
Yea, and I did run into that. Oh well, bummer.

nknapp:
JSON.parse also does not support functions. It would also not fix the vulnerabilities because the prototype functions would still be there.

I actually would like to do the white listing differently in 5.0. It would probably make more sense to pass class constructors as a runtime option and allow prototype properties of exactly those classes.

tmaiaroto:
Hmm, not in front of computer, but you sure the prototype functions would still be there? Because I was able to pass an object that way and it went through. So if it still carries a vulnerability, it's getting through the check.

PR #1635, opened 2020-01-09
feat: default options for controlling proto access

nknapp:
feat: default options for controlling proto access

This commmit adds the runtime options 
- `allowProtoPropertiesByDefault` (boolean, default: false) and 
- `allowProtoMethodsByDefault` (boolean, default: false)`
which can be used to allow access to prototype properties and
functions in general.

Specific properties and methods can still be disabled from access
via `allowedProtoProperties` and `allowedProtoMethods` by 
setting the corresponding values to false.

The methods `constructor`, `__defineGetter__`, `__defineSetter__`, `__lookupGetter__`
and the property `__proto__` will be disabled, even if the allow...ByDefault-options
are set to true. In order to allow access to those properties and methods, they have
to be explicitly set to true in the 'allowedProto...'-options.

A warning is logged when the a proto-access it attempted and denied 
by default (i.e. if no option is set by the user to make the access
decision explicit)


Gerrit0:
:+1: thanks! Looks like this is a good balance between protecting users by default and allowing them the flexibility to continue using prototype methods in projects where that is necessary.

PR #1657, opened 2020-03-04
docs: fix release notes comparison link

skeggse:
The link here isn't right, which makes spelunking a little harder.


nknapp:
Thanks

PR #1662, opened 2020-03-19
fix(4.x): migrate from optimist to yargs

AviVahl:
closes #1658
closes #1661

adapted code from `master` to latest `yargs` (`.option` calls).

*4.x*:
`found 188 vulnerabilities (169 low, 4 moderate, 14 high, 1 critical) in 5815 scanned packages`

*4.x with this PR*:
`found 32 vulnerabilities (17 low, 1 moderate, 13 high, 1 critical) in 5829 scanned packages`

PR #1666, opened 2020-03-29
fix(4.x): migrate from optimist to yargs

### LINKED ISSUE #1658: Use yargs instead of deprecated optimist in 4.x release

We are using the latest `handlebars 4.7.3`, but we noticed that one of the dependencies `optimist` is deprecated. There is a ticket to switch to `yargs` instead (https://github.com/wycats/handlebars.js/issues/1179) and it was closed with the merged PR (https://github.com/wycats/handlebars.js/pull/1180). However, the merged PR is not in 4.x release yet. Please fix this in 4.x release.

### LINKED ISSUE #1661: Peer Dependency has Security Vulnerability
Hello there,

since today, a new issue in minimist < 1.2.3 is found:
https://www.npmjs.com/advisories/1179

This lib is using optimist 0.6.1, which requires minimist in 0.0.10, which is obviously affected.
Will there be a new version with a fix for this? Our Security-Scanner is running wild because of this.

Thanks & Best Regards

fabb:
Follow-up on #1662, added a few unit test on top of the original PR.

closes #1658
closes #1661

Note that there are not yet tests for all command line options. I know too little about handlebars yet (it's just a sub-sub-dependency of a package I use). If someone has good suggestions on what makes sense to test, fire away, I will update the PR.

Since I use `npm audit` to fail my build to not get vulnerable packages get released to production, I'm very interested into getting this merged.

Tested:
- [x] -f
- [x] --map
- [x] -a
- [x] -c
- [x] -h
- [x] -k
- [x] -o
- [x] -m
- [x] -n
- [x] -s
- [x] -N
- [x] -i
- [x] -r
- [x] -p
- [x] -d
- [x] -e
- [x] -b
- [x] -v
- [x] --help


aorinevo:
I recommend looking at yargs unit tests and maybe mimicking what is done there to test handlebars cli options.

@fabb, happy to help on this, can I get permissions to your fork?

aorinevo:
Seems like the unit testing structure in place will make this some what difficult.

Will try to leverage the current unit testing structure to get this done.

fabb:
@aorinevo I invited you to this repo, feel free to commit to this branch, thanks for offering! I‘ll probably only get to continue on this in 2 days. I‘ll then have a look at what you‘ve done in the mean time, and at the yargs tests as you suggested. I guess for coordination it should be enough when we write here in this PR when we start and stop making changes, what do you think?

aorinevo:
I think that makes sense.  I'll update with comments accordingly.

aorinevo:
Tracking testing progress on my end via:

- [ ]  -f
- [ ]  --map
- [ ]  -a
- [ ]  -c
- [ ]  -h
- [ ]  -k
- [ ]  -o
- [ ]  -m
- [ ]  -n
- [ ]  -s
- [ ]  -N
- [ ]  -i
- [ ]  -r
- [ ]  -p
- [ ]  -d
- [ ]  -e
- [ ]  -b
- [ ]  -v
- [ ]  --help

aorinevo:
@fabb, please give me a heads up before merging any changes from your end on this branch.  I'd like to squash my commits into one. It's easier if there are no interwoven commits from other contributors.

aorinevo:
@fabb, unfortunately I squashed your commits down with mine.  

If you have those commits locally, I can add them back. Just push them to a branch called `faab-mty-commits` and I'll take it from there.

aorinevo:
Added back @AviVahl's commits.

aorinevo:
Going to pause here for a bit.  

Added unit tests for 8 options.  Might pick this up again later.

fabb:
> @fabb, unfortunately I squashed your commits down with mine.
> If you have those commits locally, I can add them back. Just push them to a branch called faab-mty-commits and I'll take it from there.

I'm not here for fame, so it's fine with me ;-)

spec/expected/version.txt
```diff
@@ -0,0 +1 @@
+4.7.3
```
fabb: maybe we can read out the dynamic value from package.json to make updating less annoying

ErisDS: That would really be preferable! It's possible to do require('path/to/package.json').version

ErisDS:
@fabb and @aorinevo Thanks so much for taking this on. I'll be supporting this PR through to release.

If anyone out there regularly uses the handlebars CLI and can verify this PR is working for them, that would be fantastic.

Otherwise I'll give this a spin once you're both happy it's done.

spec/expected/version.txt
aorinevo: Might require reworking the task runner but definitely doable and a better approach.

aorinevo:
Thanks @ErisDS, I’ll try to get this wrapped up tonight :)

spec/expected/version.txt
ErisDS: another approach might be just verifying that the output is `/\d+\.\d+\.\d+/` but I wouldn't worry about missing a test for this one command if it's too fiddly. 

Main thing is that having this fail after every release would be a major maintenance burden 😅

aorinevo:
@fabb, picking this backup now. Let me know if you have any WIP.

aorinevo:
@ErisDS, I'd like to _recommend_ skipping a unit test for `-b` as use of BOMS is not required or recommended (see https://en.wikipedia.org/wiki/Byte_order_mark).  

Thoughts?

spec/expected/version.txt
aorinevo: Like most things, this ended up being pretty straight forward 🥳.

aorinevo:
@ErisDS, assuming we skip `-b`, this PR is ready for review.

aorinevo:
Added last remaining unit test for `-b`.  

Unit test is valid as it runs against a file that includes a BOM (see image below).

Screen Shot 2020-03-30 at 8 15 53 PM

DeeDeeG:
_(Disclaimer: This is just something I tried in the volunteer spirit, while we wait for a proper review. I am not fluent with JS projects, so take this with a grain of salt/feel free to ignore if it is not helpful. Apologies in advance if that is the case.)_

Should these tests pass with the old `optimist`-based code dropped in place of the new `yargs` code? I tried to do this (new tests, old code/dependencies) and it failed, but it may have been a mistake on my end.

(My Travis runs can be viewed here: https://travis-ci.com/github/DeeDeeG/handlebars.js/branches)

spec/expected/version.txt
aorinevo: This file can be removed now.

aorinevo:
@DeeDeeG, I'm rather curious myself why that's the case.  I'm going to take a look.

aorinevo:
@DeeDeeG, still looking for the root cause but worth noting that all the tests pass with the exception of the unit test for `--help` when I use `optimist`.

aorinevo:
Found it :)

Optimist puts the `--help` option at the bottom of the menu while `yargs` places it at the top.

fabb:
Oh wow, I thought there would be some leftovers for me when I continue today in the evening 😅
Thanks for the work you put in!

tasks/test-bin.js
```diff
@@ -7,22 +7,213 @@ const chai = require('chai');
...
+  },
+  {
+    binInputParameters: ['--help'],
+    outputLocation: 'stdout',
+    expectedOutputSpec: './spec/expected/help.menu.txt'
+  },
+  {
+    binInputParameters: ['-v'],
+    outputLocation: 'stdout',
+    expectedOutput: require('../package.json').version
```
fabb: 👍

fabb:
LGTM!

i think the `--help` test is a bit of overkill, but it can easily be removed if it breaks too often in future PRs.

ErisDS:
@DeeDeeG your suggestion was really useful - great idea thanks :) 

@fabb and @aorinevo seems like this is ready to go? I'll take it for a spin in a bit.

aorinevo:
@ErisDS, yep, all done :)

ErisDS:
Manually merged via b440c38886f62bd1801b3bb2da78f1f341c2b121, d79212a662d2aae106449338a28717352ec65ce6 and 80c4516fdad5d8fcb6e24faca51db97bd6cc94c5

fabb:
Thanks everyone!

aorinevo:
Fancy :)

Thank you @ErisDS, @fabb, @AviVahl, and @nknapp!

AviVahl:
Thanks all for taking this across the finishing line.

@ErisDS any chance this could be tagged and published? :)

DeeDeeG:
By the way... I can confirm that, with `optimist` plunked in instead of `yargs`, all these tests _(other than the `--help` test, as discussed above)_ are passing.

Here is a green/passing CI run (with `optimist` instead of `yargs`, and the `--help` test disabled): https://travis-ci.com/github/DeeDeeG/handlebars.js/builds/157285103

ErisDS:
I have been working on the release since I merged the changes 😉 

Unfortunately, I've hit several road blocks. The latest one is that the sourcemap tests [here](https://github.com/wycats/handlebars.js/pull/1666/files#diff-44e0212ed1b982b5e5cec6028eedd4b4R150-R176) result in the file source.map.amd.txt being written to. The content doesn't change but the timestamp on the file does.

The release process runs `git diff-index --name-only HEAD --` which then fails and blocks the release.

I'm looking to see if I can find a fix, but I'm also wondering why that test is there twice. Is it a typo?

DeeDeeG:
@ErisDS looking into that release process snag: I noticed just running the command the test runs will also write to that sourcemap file. (Minimal-ish reproducer for the issue).

`handlebars -i '<div>1</div>' -a -m -N test --map './spec/artifacts/source.map.amd.txt'`
(Updates the modification date for `./spec/artifacts/source.map.amd.txt`)

I'm not sure if there's a way to really test/run the feature without updating [the modification date of] the file.

A quick workaround would be:

`git checkout HEAD -- spec/artifacts/source.map.amd.txt`
(Freshly writes the file out to disk from the `git` index, apparently resetting `git`'s expectation of the modification date to "now" in the process)

Kinda hacky, but it works? (To be run after the test and before `git diff-index --name-only HEAD --`)

AviVahl:
Should probably tell the cli test to output to a temp file at [os.tmpdir()](https://nodejs.org/api/os.html#os_os_tmpdir) and delete it afterwards (instead of trying reset a git-tracked file).

Could use [fs.mkdtempSync()](https://nodejs.org/api/fs.html#fs_fs_mkdtempsync_prefix_options) to get a unique folder inside temp for each run.

DeeDeeG:
I don't think the test itself writes anything, other than via running `handlebars`.

I may be wrong, but I think `handlebars` is writing to a file even though there are no changes to be made. Perhaps `handlebars` needs to skip writing out the sourcemap file to disk if there are no changes to be made.

nknapp:
In other projects, I use a folder "test-tmp/test-name" to write files generated by the tests. The advantage is that the files are easier to find and inspect if a test fails. "test-tmp" is part of gitignore and the subfolder is deleted and recreated *before* the test runs.

It may indeed be the case that all tests rib twice on release, but I am not sure. I think the unit tests do, for different environments.

aorinevo:
> I may be wrong, but I think `handlebars` is writing to a file even though there are no changes to be made. Perhaps `handlebars` needs to skip writing out the sourcemap file to disk if there are no changes to be made.

I can confirm that that is indeed what handlebars does.

aorinevo:
@ErisDS, happy to take this on.  I think I see a fix for this with a low-level of effort; however, I don't want to duplicate effort if you're already on this.

Let me know :)

aorinevo:
The fix is to use unlink.  That will remove the file after the test has been run.

@ErisDS, I'll prepare a PR for this.

aorinevo:
For your reviewing please :) https://github.com/wycats/handlebars.js/pull/1667 (cc @ErisDS)

ErisDS:
@DeeDeeG have cherry picked your commit removing the duplicate

I believe [Nils' suggestion](https://github.com/wycats/handlebars.js/pull/1666#issuecomment-606879693) is the best approach for a fix here.

ErisDS:
I've used the tmp dir approach to fix this locally and am going to push it to unblock the release.

DeeDeeG:
Thank you everyone for your efforts. This was a productive and cordial experience, and I was glad to be a part of it. Open-source can be pretty cool sometimes.

Hope everyone is safe and well out there.

ErisDS:
Hey everyone 👋  The release of 4.7.4 is now fully complete, including the various packages and updating the handlebars website. 

Thank you to everyone who helped get this over the line, really appreciate it ❤️

PR #1667, opened 2020-03-31
fix: file creation bug on tests.

aorinevo:
# Changes
- When running `grunt test:bin` some files are created that fail CI if not removed.  For each test, if a file is created it is cleaned up after the test run in complete.
- Removed duplicate test.
- Reused an existing artifact thereby allowing for deletion of one artifact file.  Updated unit tests accordingly.

# Notes
- Fixes https://github.com/wycats/handlebars.js/pull/1666#issuecomment-606805248


spec/expected/non.default.extension.amd.js
```diff
@@ -1,6 +1,15 @@
 define(['handlebars.runtime'], function(Handlebars) {
     Handlebars = Handlebars["default"];  var template = Handlebars.template, templates = Handlebars.templates = Handlebars.templates || {};
-  return templates['non.default.extension'] = template({"compiler":[8,">= 4.3.0"],"main":function(container,depth0,helpers,partials,data) {
-      return "<div>This is a test</div>";
+  return templates['example_2'] = template({"compiler":[8,">= 4.3.0"],"main":function(container,depth0,helpers,partials,data) {
```
aorinevo, line 3: This change is due to reusing an artifact.

spec/precompiler.js
```diff
@@ -307,7 +307,7 @@ describe('precompiler', function() {
       Precompiler.loadTemplates(
         { files: [__dirname + '/artifacts'], extension: 'hbs' },
         function(err, opts) {
-          equal(opts.templates.length, 2);
+          equal(opts.templates.length, 1);
```
aorinevo, line 310: As `non.default.extension.amd.hbs` artifact was removed in this PR, the count is now back to 1.

DeeDeeG:
FWIW, I ran `npm run test` on this branch, and then `git diff-index --name-only HEAD --`.

The tests all passed as expected, and the `git diff-index` command turned up clean.

ErisDS:
Hmm, I think this change is a little confusing and changes what is being tested.

In the original test:

```
{
    binInputParameters: [
      '-i',
      '<div>1</div>',
      '-a',
      '-m',
      '-N',
      'test',
      '--map',
      './spec/artifacts/source.map.amd.txt'
    ],
    outputLocation: 'stdout',
    expectedOutputSpec: './spec/expected/source.map.amd.js'
  },
```

`./spec/artifacts/source.map.amd.txt`

Contains the sourcemap:

```
{"version":3,"sources":["test"],"names":["compiler","main","container","depth0","helpers","partials","data","useData"],"mappings":"kHAAA,CAAAA,SAAA,CAAA,EAAA,YAAAC,KAAA,SAAAC,EAAAC,EAAAC,EAAAC,EAAAC,GAAA,MAAA,gBAAAC,SAAA"}
```

And `spec/expected/source.map.amd.js` contains a compiled handlebars template with a sourcemap ref:

```
define(["handlebars.runtime"],function(e){var t=(e=e.default).template;return(e.templates=e.templates||{}).test=t({compiler:[8,">= 4.3.0"],main:function(e,t,a,n,i){return"<div>1</div>"},useData:!0})});
//# sourceMappingURL=./spec/artifacts/source.map.amd.txt
```

With the new test:

```
 {
    binInputParameters: [
      '-i',
      '<div>1</div>',
      '-a',
      '-m',
      '-N',
      'test',
      '--map',
      './spec/artifacts/source.map.amd.json'
    ],
    outputLocation: './spec/artifacts/source.map.amd.json',
    expectedOutputSpec: './spec/expected/source.map.amd.json'
  },
```

We're now testing the sourcemap has the correct config, not the template. 

That might well be a valid test, but it's a different test to the one that was there before 🙈 

Nothing to do with this PR,  `outputLocation` declares where the test should look for its output, rather than changing where the output goes, which I also find a bit confusing.

aorinevo:
> We're now testing the sourcemap has the correct config, not the template.

This is a bi-product of passing a string via cli to the the handlebars precompiler vs. reading from a template file.

> We're now testing the sourcemap has the correct config, not the template.

IMO, this is a better unit test than the previous one as it asserts the content of the source map.  We can improve on this by adding another assertion to test that the source map is correctly referenced within the compiled template.  This will require us to switch from passing a string to the complier via the cli to reading from a template file.

Let me know your thoughts.  Happy to close the gap as per above comments.

ErisDS:
> IMO, this is a better unit test than the previous one as it asserts the content of the source map. 

I disagree 🙂 Handlebars isn't responsible for the content of the source map being correct, that's the job of the source-map module.

The original test checked that the handlebars template had the correct sourcemap string, which is the behaviour specific to handlebars.

A second thing to test would be that the linked source map file actually exists, but I see no reason to verify the contents.

TBH I think the correct fix for this issue is [Nils' suggestion](https://github.com/wycats/handlebars.js/pull/1666#issuecomment-606879693).

ErisDS:
I've used the tmp dir approach to fix this locally and am going to push it to unblock the release.

If you would like to update this PR to add an additional test to check that the sourcemap file is created, and make the other modifications here, will merge after release :)

aorinevo:
> I disagree 🙂 Handlebars isn't responsible for the content of the source map being correct, that's the job of the source-map module.

Quiet right; I'm convinced :)

I'll update the tests accordingly.

nknapp:
Do we still need this PR or can it be closed?

aorinevo:
Let's keep this open as until I've had a chance to review what we landed in `15.x` as an alternative to this change.

I hope to get it this weekend.

jaylinski:
@aorinevo Is this still relevant?

tasks/test-bin.js
```diff
@@ -155,24 +150,10 @@ const testCases = [
       '-N',
       'test',
       '--map',
-      './spec/artifacts/source.map.amd.txt'
```
jaylinski, line 158: Was already deleted in e46baa1fe23b3a1bee07b085e48700c090218b25.

jaylinski:
CI doesn't seem to fail, we'll leave the duplicate test for now.

PR #1669, opened 2020-04-02
test: fix integration tests

nknapp:
Ensure that integration tests fail, if a single test fails
The build for this PR will fail, if the tests are fixed correctly, because the `yargs` upgrade is incompatible to older node versions.


ErisDS:
Merging this as it demonstrates that the build is broken, and that subsequent changes will fix it

Issue #1668, opened 2020-04-01
What is the Node.js version requirement for 4.x?


dougwilson:
First, I want to start off with a thanks for all the work you guys do.

I have a question that came up after I tried upgrading to the latest 4.x patch version of 4.7.4: what Node.js versions does this module actually support? This is because suddenly it seems there is a dependency in `handlebars` that requires Node.js 8+, but the declaration in `handlebars` is still quite low, at Node.js 0.4.7+ (https://github.com/wycats/handlebars.js/blob/v4.7.4/package.json#L21).

Now, I certainly have not tried to run this module with Node.js 0.4, haha, but I seem to be suddenly having CI failures with Node.js 6, related to the `yargs` package.

bcoe:
@ErisDS hey👋 I'm the maintainer of yargs. I saw the recent handlebars update that migrates from optimist to yargs (which is exciting) ... but, yargs@15.x.x requires Node 6, so might be breaking for some handlebars users (based on your engines field).

yargs@15 should work for Node.js 6 (we haven't dropped Node 8 yet).

What are the version support goals for handlebars, I could work with you to target a slightly older version of yargs (but can't support all the way back to 0.4).

I might just be tempted to call this a major change for people.

nknapp:
There are actually tests that should have failed in this case, but even though I can see the error messages hidden in the Travis log, the build itself did not fail.

The test tries to use the handlebars CLI with node versions starting at 0.10.

@ErisDS I didn't see this coming. Shall we revert?

bcoe:
@nknapp @ErisDS I think it could be fairly disruptive to folks if you don't revert, what if you reverted, and made it a `major` with an engines update?

You could install `yargs@14` which is not vulnerable, and supports back to Node 6.... But both Node 6 and 8 are EOL, and will no longer take patches. So I would put some thought into what your version support goals are (0.4 is unlikely to be able to be targeted by a patched version of optimist or yargs without a good chunk of work).

nknapp:
@ErisDS I have fixed the integration tests in #1669

ErisDS:
@dougwilson thanks for bringing this to our attention ❤️ Could you explain where and why you're running against Node v6 just so we have a better understanding?

Me and @nknapp have had a chat and think that it's time for Handlebars' extensive Node compatibility  to end as there aren't enough maintainers and quite frankly, I doubt there's any need for support < v6.

[...]

PR #1671, opened 2020-04-02
docs: add Node.js support policy

nknapp:
Initiated by #1668 we had a discussion about support for Node.js versions. This is the write-up of result.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -79,7 +79,7 @@ Supported Environments
 
 Handlebars has been designed to work in any ECMAScript 3 environment. This includes
 
-- Node.js v6+
+- Node.js (see below)
 - Chrome
 - Firefox
 - Safari 5+
@@ -90,6 +90,17 @@ Older versions and other runtimes are likely to work but have not been formally
 tested. The compiler requires `JSON.stringify` to be implemented natively or via a polyfill. If using the precompiler this is not necessary.
 
 [![Selenium Test Status](https://saucelabs.com/browser-matrix/handlebars.svg)](https://saucelabs.com/u/handlebars)
+
+## NodeJS support policy
+
+Handlebars aims to support the **current** version of Node.js as well as **active LTS** versions.
+
+Our policy regarding Node.js is that dropping support for unsupported Node.js-versions 
+is **not** considered a breaking change. Please have a look at the Node.js support timetable
+at https://github.com/nodejs/Release/blob/master/README.md.
+
+This policy does not affect browser compatibility. Dropping support for a currently supported 
+browser version will be a breaking change and a major version bump.
 
 Performance
 -----------


bcoe:
Hey @nknapp @ErisDS, thought you might find this interesting reading (It's the version support goals I've been drafting for Node.js libraries at Google):

https://github.com/googleapis/google-cloud-node#supported-nodejs-versions

I'm working on a more thorough blog post I'll share pulling together my thoughts. It's motivated by this exact challenge I've been watching handlebars deal with this week.

Appreciate your hard work, and that open source can sometimes be incredibly frustrating 😆 

Good thinking calling out your goals explicitly.

bcoe:
The opinion I've developed, as a library maintainer, is that 30 months is reasonable amount of time to support a Node.js version (which is the LTS length). It creates a burden for maintainers to support EOL versions:

* making it difficult to take patches from libraries in the dependency tree.
* opening you up to security vulnerabilities.
* making it so you can't take advantage of new runtime features.

At the same time, I do strongly agree that dropping a Node.js version should be a major bump. Otherwise I think you'll have the unintended effect that people start pinning to specific versions of handlebars.

I'm very sympathetic to handlebars' stance around dealing with the optimist vulnerability.

nknapp:
@bcoe just wanted to say that I see your point and will think about it. 

I will try to briefly make my point as well: 

I belief that major version bumps should be never be done lightly, avoided if possible and carefully planned if necessary. `4.x` now exists since 2015. Since then, we have mostly done bugfixes, added a few options and fixed security issues. But we haven't had any breaking changes in the documented API, so there was no need to release 5.0

Node.js drops support for an LTS version every year. If a security issue can only be resolved by upgrading a dependency to a new version and if that new version does not support the dropped Node.js version anymore, then we need to release a new major version of Handlebars, just because dropping support for the old version, even though we had no other breaking changes.

Also, only a very small part of Handlebars depends on the version of node.js (only the precompiler). Most of the code (compiler, runtime) is transpiled at the moment and does not depend on any node.js API. 

So in the worst case, we have one major release every year with no other breaking changes. Maybe with only security fixes. This may well happen at our current speed of development and it really bothers me to have to do that. 

On the other hand I can understand the benefits of your point of view. As I said, I am thinking about it, and maybe we should discuss this again.

jaylinski:
@nknapp I think we should merge this into `master` (v5), but keep v4 Node.js support as is.

akarsh (approved):

akarsh:
Hello @nknapp 

Good PR 👍

I agree with the PR changes as the workflow file https://github.com/handlebars-lang/handlebars.js/blob/master/.github/workflows/ci.yml#L20 mentions the minimum supported Node.js version is currently 18.x and Node.js projects depenending on GitHub Actions workflows are forced by GitHub Actions to updrade their projects to the supported Node.js version in https://github.com/actions/setup-node/

nknapp:
It's been a long time since I  opened this MR and I would like to revert my statement. It should be a breaking change. 
The only reason to not run tests against all node versions it lacking support of the test-frameworks as far as I can remember.

I will close this PR now. 

@jaylinski maybe a pragrmatic approach to dropping the currently old node versions is to discard that changes that have been done one the master branch and start clean off 4.x to get a 5.0, except for the technical changes like using the external parser repo maybe.

I don't know how much work you have put into the new version already, but all the features in there have been lying around for 8 years now without a release. I doubt that anybody really needs them.

PR #1672, opened 2020-04-03
Switch cmd parser to latest minimist

dougwilson:
This pull request just swaps the command line parser to the latest version of minimist, which removes a couple of "regressions" (like the changed description of version and the changed position of the help option), but mainly both (1) removes the minimist security notice and (2) keeps the 4.x on the same Node.js version support plan.

I hope this pull request can be considered.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -79,7 +79,7 @@ Supported Environments
 
 Handlebars has been designed to work in any ECMAScript 3 environment. This includes
 
-- Node.js v6+
+- Node.js
 - Chrome
 - Firefox
 - Safari 5+
--- a/bin/.eslintrc.js
+++ b/bin/.eslintrc.js
@@ -1,0 +1,6 @@
+module.exports = {
+  rules: {
+    'no-console': 0,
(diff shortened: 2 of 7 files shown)


dougwilson:
Travis CI status did not update on the pull request, but here is a link to the passing Travis CI build: https://travis-ci.org/github/wycats/handlebars.js/builds/670392714

ErisDS:
Will take a look ASAP

aorinevo:
Happy to take this out for a spin :)

dougwilson:
This also fixes #1673

ErisDS:
Just to let you know, I am looking to merge and release this.  I would prefer to do the node changes separately as a revert though. 

I then plan to revert this change on the master branch because (as you said on twitter) the yargs changes are slightly more maintainable long term.

This would leave us with #1673 on master, but I've found a fix for that.

ErisDS:
I've pushed the change to split out the node changes to the PR which also has the side effect of getting the travis build to show here properly.

PR #1675, opened 2020-04-08
docs: Fix simple typo, mulitple -> multiple

timgates42:
There is a small typo in lib/handlebars/compiler/whitespace-control.js.

Should read `multiple` rather than `mulitple`.


nknapp:
Thanks for that correction. I actually thought I had already merged it. But I must have missed it somehow.

Issue #1676, opened 2020-04-11
Dependency Management Automation


aorinevo:
_Note: this issue is a port of https://github.com/NerdWalletOSS/shepherd/issues/111_ 

## Narrative

As a maintainer of Handlebars, I want a quick and easy way to keep dependencies up-to-date and reduce the level-of-effort associated with the corresponding releases.  

## Proposed Change

The proposed change is to integrate renovate, renovate approve, semantic commits, and semantic-releases.  

With the right combination of settings across these four Github apps, CI/CD can take care of updating minor and patch dependencies (relative to the semantic commit message), approving and merging of the corresponding PRs, publishing a new tag version, and finally updating release notes (if needed) and changelog.

## ROI
- near _real time_ vulnerability and security patches
- reduced overhead for maintainers
- reduced CI/CD failures for consumers (at least for those that run `npm audit` as part of their CI)

## DependaBot vs RenovateBot

| |  DependaBot | RenovateBot |
|----|----|----|
|post install config (i.e. run `npm dedupe`) | No | Yes |
|price | Free | Free |

## Nice to have
- WIP app

## Notes
- Shepherd currently uses dependabot (see https://dependabot.com)
- Renovate Approve (see https://github.com/apps/renovate-approve)
- Renovate (see https://github.com/marketplace/renovate)
- Semantic Release (see https://github.com/semantic-release/github)
- Semantic Pull Requests (see https://github.com/zeke/semantic-pull-requests)

ErisDS:
Just want to add a +1 for renovate here. I’ve used it extensively and it is a fantastic tool.

nknapp:
do we need extra config for renovate in the repo. If yes, we should add it first and install the bot.

aorinevo:
We will need a `renovate.json` file to house the configurations however it is worth noting that renovate can be enabled with without risk:

> Renovate will not make any changes to your repository or raise any further Pull Requests until after you merge this initial Pull Request. So if there is anything about the Pull Request that you don't like or understand, take your time to read documentation or ask questions in one of our support forums and merge the PR only once you're satisfied with the result. You can edit your Renovate configuration within this renovate/configure branch and Renovate will keep updating the description in the PR to match, so you can keep doing that until you're satisfied with the results.

~ from https://docs.renovatebot.com/configure-renovate/

jaylinski:
[...]

PR #1677, opened 2020-04-11
[WIP] feat: add dependency management automation (#1676).

aorinevo:
## Changes
- See Issue #1676


nknapp:
Thanks for this. It would be great to use semantic-release to do the actual release. It would simplify many things. 

I think we should approach this in the following way: 

1. Add the semantic-release config to the `4.x`-branch, but not the github-action. For the time being, we can run `npx semantic-release` locally but we can still revert to the old way of using `yo release` if it does not work as expected.
2. Migrate travis and appveyor to github actions.
3. Once we've confirmed that `npx semantic-release` covers the current release workflow and all targets, we can move it into the action.

Here are my pain points:

* I am not sure that I know everything that `yo release` does in the background. I also don't know yet, what tasks/scripts `npx semantic-release` actually executes in order to build the project. But I am pretty sure that those are not the same at the moment. I want to test this manually before getting into automation.
* My `npm` and `gem` accounts have MFA activated, which means that I have to enter an one-time-password when I publish packages. This is for a good reason: To prevent that something like the [eslint-scope issue](https://eslint.org/blog/2018/07/postmortem-for-malicious-package-publishes) happens to Handlebars or other packages that I publish on npm. If we fully automate releases, I would like to maintain that level of security. I think we first have to explore the possibilities here.

I think it is acceptable to type `npx semantic-release` in order to get a new release. And it's a major improvement compared to what we have to do right now.

jaylinski:
It would be nice if we could do npm package provenance for verified builds:

https://github.blog/2023-04-19-introducing-npm-package-provenance/

PR #1678, opened 2020-04-14
Make ` delimit string literals

brabeji:
I would like to allow my users to write handlebars templates with complex string literals in helper params, containing lots of `"`s and `'`s without the need to escape them. I see that `` ` `` is a reserved control character, which is awesome, but I didn't see its use in current docs. Is this something that can be added to `v5`?

Thanks!

### DIFF
--- a/spec/tokenizer.js
+++ b/spec/tokenizer.js
@@ -453,6 +453,12 @@ describe('Tokenizer', function() {
     shouldBeToken(result[3], 'STRING', 'baz');
   });
 
+  it('tokenizes mustaches with String params using back-ticks as "OPEN ID ID STRING CLOSE"', function() {
+    var result = tokenize('{{ foo bar `baz` }}');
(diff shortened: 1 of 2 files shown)


nknapp:
You can also use `[string "literal" with 'quotes']` in templates. Isn't that enough?

brabeji:
@nknapp That would be sufficient, I don't care about the delimiter being exactly `` ` ``. 
Unfortunately, I need it to work as a helper parameter which it doesn't seem to. Is `[]` a v5 feature by chance? It seems to always evaluate to undefined in v4

[https://codesandbox.io/s/inspiring-shape-8oy4x?file=/src/index.ts](https://codesandbox.io/s/inspiring-shape-8oy4x?file=/src/index.ts)

nknapp:
Right, brackets are used for special chars in path expressions only. I confused the two.

No, we haven't planned anything like that at the moment and it falls into the category "language feature", which we actually don't want to do right now until we have a formal language spec #1277

brabeji:
That makes sense, thank you for the feedback. Closing for now.

Issue #1418, opened 2018-01-10
Support for maps, sets, and custom iterables in built-in "each" helper?


jasonh-brimar:
When using Handlebars in an ES6 environment, the built-in `each` helper's limitation of supporting only arrays and generic objects becomes inconvenient. To work around this, I started registering my own version of the `each` helper that supports arrays, maps, sets, custom iterables, and generic objects. That helper is below.

Is there a plan or willingness to introduce support for these types of lists in the built-in `each` helper? I ask because I understand that Handlebars aims to avoid polyfills and I imagine that the only way of making the new helper work without compromising on browser support would be to progressively enable support for the different list types dependent on the environment's native or pre-polyfilled support for `Set`, `Map`, and `Symbol`.

```js
Handlebars.registerHelper("each", function (contexts, options) {

    // Throw a runtime exception if options were not supplied.
    if (!options) {
        throw new Handlebars.Exception("Must pass iterator to #each");
    }

    // If the "list of contexts" is a function, execute it to get the actual list of contexts.
    if (typeof contexts === "function") {
        contexts = contexts.call(this);
    }

    // If data was supplied, frame it.
    const data = options.data ? Object.assign({}, options.data, { _parent: options.data }) : undefined;

    // Create the string into which the contexts will be handled and returned.
    let string = "";

    // Create a flag indicating whether or not string building has begun.
    let stringExtensionStarted = false;

    // Create a variable to hold the context to use during the next string extension. This is done to
    // allow iteration through the supplied list of contexts one step out of sync as they are looped
    // through later in this helper, ensuring a predictable sequence of value retrieval, string
    // extension, value retrieval, string extension...
    let nextContext;

    // Create a function responsible for expanding the string.
    const extendString = (final = false) => {

        // If other contexts have been encountered...
        if (nextContext) {

            // Expand the string using the block function.
            string += options.fn(nextContext.value, {
                data: data ? Object.assign(data, {
                    index: nextContext.index,
                    key: nextContext.key,
                    first: !stringExtensionStarted,
                    last: final
                }) : undefined,
                blockParams: [nextContext.key, nextContext.value]
            });

            // Note that string extension has begun.
            stringExtensionStarted = true;

        // If no contexts have been encountered and this is the final extension...
        } else if (final) {

            // Expand the string using the "else" block function.
            string += options.inverse(this);

        }

    };

    // If a list of contexts was supplied...
    if (contexts !== null && typeof contexts !== "undefined") {

[...]

PR #1679, opened 2020-04-16
Support Map for iteration and nested paths

karlvr:
Re #1418 and #1557 this PR adds support for accessing properties in Maps, and iterating over the values in Maps using `#each`.

e.g.

```javascript
var foo = new Map();
foo.set('expression', 'beautiful');
...
```
`+`
```handlebars
Goodbye {{foo/expression}} world!
{{#each foo}}
The {{@key}} is {{.}}.
{{/each}}
```
`=`
```
Goodbye beautiful world!
The expression is beautiful.
```

- [X] Please don't start pull requests for security issues. Instead, file a report at https://www.npmjs.com/advisories/report?package=handlebars
- [X] Maintain the existing code style
- [X] Are focused on a single change (i.e. avoid large refactoring or style adjustments in untouched code if not the primary goal of the pull request)
- [X] Have good commit messages
- [X] Have tests
- [X] Have the [typings](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html) (lib/handlebars.d.ts) updated on every API change. If you need help, updating those, please mention that in the PR description.
- [X] Don't significantly decrease the current code coverage (see coverage/lcov-report/index.html)
- [X] Currently, the `4.x`-branch contains the latest version. Please target that branch in the PR.

### DIFF
--- a/lib/handlebars/helpers/each.js
+++ b/lib/handlebars/helpers/each.js
@@ -3,7 +3,8 @@ import {
   blockParams,
   createFrame,
   isArray,
-  isFunction
+  isFunction,
+  isMap
 } from '../utils';
 import Exception from '../exception';
 
@@ -60,7 +61,7 @@ export default function(instance) {
             execIteration(i, context[i], i, i === context.length - 1);
           }
         }
-      } else if (context instanceof Map) {
+      } else if (isMap(context)) {
         const j = context.size - 1;
         for (const [key, value] of context) {
           execIteration(key, value, i, i === j);
--- a/lib/handlebars/helpers/each.js
+++ b/lib/handlebars/helpers/each.js
@@ -33,7 +33,7 @@ export default function(instance) {
       data = createFrame(options.data);
     }
 
-    function execIteration(field, index, last) {
+    function execIteration(field, value, index, last) {
       if (data) {
         data.key = field;
         data.index = index;
@@ -47,12 +47,9 @@ export default function(instance) {
 
       ret =
         ret +
-        fn(context[field], {
+        fn(value, {
           data: data,
-          blockParams: blockParams(
-            [context[field], field],
-            [contextPath + field, null]
-          )
+          blockParams: blockParams([value, field], [contextPath + field, null])
         });
     }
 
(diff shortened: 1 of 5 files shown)


karlvr:
I was trying to fix the change in code formatting around the `blockParams`, but it looks like there's some automatic reformatting on commit? Something's doing something :-D

karlvr:
I'm sorry I didn't notice the failing tests outside of the node environment before. I am trying to work out what's going wrong now... I think perhaps it's a product of webpack.

karlvr:
That's the ticket. I'm keen to discuss; Handlebars is central to my project and I'm trying to make the decision between `Map` and POJOs.

nknapp:
Thanks for submitting this PR. I haven't been able to answer earlier, but I think your PR makes sense. There are just one point that I would like to discuss and one change request. I haven't had a very close look at the code yet though.

## Is this a breaking change?

I am not sure if this change should go into the `4.x`-branch. In #1557, you said that Maps could be iterated at the moment, they would be iterate like an array of key-value pairs, rather than like an object (values with `@key` containing the key).

I would like to hear some opinions (@wycats, @ErisDS, @aorinevo) if this should be considered a breaking change. If we interprete semver strictly, it is not breaking, because iterating Maps is not documented anywhere.

But one could also argue that some people might already use `{{#each}}` to iterate maps and expect key-value pairs in the iteration. Support for iterable objects was added only a couple of months ago in [4.4.0](https://github.com/handlebars-lang/handlebars.js/blob/master/release-notes.md#v440---september-29th-2019), so there is only a limited changes that this is the case

## Tests

There is a new way of writing tests (see #1683 for details). You IDE probably didn't show the `shouldCompileTo`-function as deprecated, but the new way is to use `expectTemplate(...).toCompileTo(...)` instead. It makes the test-code much more readable.

I am also not sure if we will get any merge conflicts with #1683, so I think it would be best to wait until that one is merged, and then rebase.

karlvr:
@nknapp thanks for your message. No, I didn't see that! I've converted those tests. I'm also happy to rebase and make any adjustments when #1683 lands.

nickshreck:
Did this ever get merged? It seems the perfect solution to handling Maps in Handlebars. I haven't found a cleaner way yet.

jaylinski:
Memo: in order to play it safe, this will be only merged in `master` (v5).

@karlvr I know it's been 3 years, but if this PR is still relevant for you, please target in on `master`. (Otherwise I'll do it myself, because I think this is a really nice feature.)

karlvr:
@jaylinski no problem! I've left things much longer than that. I actually didn't end up using this. I had a quick look and it looks like it might benefit from a little insight to rebase it (particularly `each.js`). If you're happy to that would be fantastic.

wind4gis:
did handlebars support map now?

jaylinski:
@karlvr I opened an updated PR in #1996. Thanks for your PR, it helped a lot!

Issue #1488, opened 2018-12-20
uglify-js is always downloaded on compatible environments


Silic0nS0ldier:
This isn't #1391 but it does have some overlap. The work done towards resolving that issue only permits functionality when `uglify-js` isn't available.

This issue here is related to how NPM treats optional dependencies. Namely, it treats them as dependencies that the package can work without, but would like to have if possible (e.g. `chokidar` has an optional dependency on `fsevents` to permit better file system event handling on unix systems).

To make `uglify-js` truly optional, it should be a `peerDependency`. This expresses compatibility without having it automatically installed, leaving the dependency up to whoever uses handlebars.

Happy to make a PR for this (including any updates that would be needed for testing, etc) if people are happy with the proposed change.

nknapp:
I'd like to see this in 5.0, because anyone who uses Handlebars in a pure Node.js-environment, actually does not need uglify at all.

Silic0nS0ldier:
I'll take a look and see if I can get a PR together then.

nknapp:
That's great. Please make sure to use the master-branch as target branch for the merge request.

Silic0nS0ldier:
PR is up. Let me know if there are any questions or concerns.

PR #1682, opened 2020-04-24
Moved `uglify-js` from `optionalDependencies` to `peerDependencies`

Silic0nS0ldier:
Addresses #1488, with this change "uglify-js" will no longer be automatically installed. When this occurs `npm install` (and some of its analogues) will log a message indicating that then peer dependency was not installed. This behaviour extends to the package with the listed peer dependency (handlebars in this case), as such "uglify-js" was also added to `devDependencies` so that it is installed for use in tests (with the same version constraint).

No new tests were created as the existing optional dependency tests already correctly covered this scenario (with the "uglify-js" missing scenario addressed via mocks).

------------------------------------------------
### PR Checklist

- [x] _(I hope)_ Have good commit messages
- [x] _(Already covered)_ Have tests
- [x] _(5.0 target, hence pointed to master)_ Currently, the `4.x`-branch contains the latest version. Please target that branch in the PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1729,7 +1729,8 @@
     "commander": {
       "version": "2.20.3",
       "resolved": "https://registry.npmjs.org/commander/-/commander-2.20.3.tgz",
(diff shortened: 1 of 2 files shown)


nknapp:
Thanks for this PR. I don't know how long it will take to publish the 5.0 release, but this definitely is a first step.

PR #1683, opened 2020-04-24
In spec tests, use expectTemplate over equals and shouldThrow

jbboehr:
So, I hate to change a bunch of code for relatively little benefit, but I have a project that exports the specification tests as JSON [[0]] and, well, it would make my life a lot easier if `expectTemplate` was used instead of `equals` or `shouldThrow` in the test suite, where possible.

I would also convert `shouldCompileTo` if you'd like, but I figured I'd keep the changes to a minimum for now.

I added two methods, `withHelpers` and `withPartials` to `HandlebarsTestBench` only because it let me port a bunch of tests with fewer changes. If you'd prefer, I would remove those and convert the tests to use `withHelper` and `withPartial` directly.

I tried to keep the changes generally linear so it would be easier to review. Let me know if there's anything you'd like changed.

[0]: https://github.com/jbboehr/handlebars-spec/

### DIFF
--- a/spec/basic.js
+++ b/spec/basic.js
@@ -6,117 +6,141 @@ beforeEach(function() {
 
 describe('basic context', function() {
   it('most basic', function() {
-    shouldCompileTo('{{foo}}', { foo: 'foo' }, 'foo');
+    expectTemplate('{{foo}}')
+      .withInput({ foo: 'foo' })
+      .toCompileTo('foo');
   });
 
   it('escaping', function() {
-    shouldCompileTo('\\{{foo}}', { foo: 'food' }, '{{foo}}');
-    shouldCompileTo('content \\{{foo}}', { foo: 'food' }, 'content {{foo}}');
-    shouldCompileTo('\\\\{{foo}}', { foo: 'food' }, '\\food');
-    shouldCompileTo('content \\\\{{foo}}', { foo: 'food' }, 'content \\food');
-    shouldCompileTo('\\\\ {{foo}}', { foo: 'food' }, '\\\\ food');
+    expectTemplate('\\{{foo}}')
+      .withInput({ foo: 'food' })
+      .toCompileTo('{{foo}}');
+    expectTemplate('content \\{{foo}}')
+      .withInput({ foo: 'food' })
+      .toCompileTo('content {{foo}}');
+    expectTemplate('\\\\{{foo}}')
+      .withInput({ foo: 'food' })
+      .toCompileTo('\\food');
+    expectTemplate('content \\\\{{foo}}')
+      .withInput({ foo: 'food' })
+      .toCompileTo('content \\food');
+    expectTemplate('\\\\ {{foo}}')
+      .withInput({ foo: 'food' })
+      .toCompileTo('\\\\ food');
   });
 
   it('compiling with a basic context', function() {
-    shouldCompileTo(
-      'Goodbye\n{{cruel}}\n{{world}}!',
-      { cruel: 'cruel', world: 'world' },
-      'Goodbye\ncruel\nworld!',
-      'It works if all the required keys are provided'
-    );
+    expectTemplate('Goodbye\n{{cruel}}\n{{world}}!')
+      .withInput({
+        cruel: 'cruel',
+        world: 'world'
+      })
+      .withMessage('It works if all the required keys are provided')
+      .toCompileTo('Goodbye\ncruel\nworld!');
   });
 
   it('compiling with a string context', function() {
-    shouldCompileTo('{{.}}{{length}}', 'bye', 'bye3');
+    expectTemplate('{{.}}{{length}}')
+      .withInput('bye')
+      .toCompileTo('bye3');
   });
 
   it('compiling with an undefined context', function() {
-    shouldCompileTo(
-      'Goodbye\n{{cruel}}\n{{world.bar}}!',
-      undefined,
-      'Goodbye\n\n!'
-    );
+    expectTemplate('Goodbye\n{{cruel}}\n{{world.bar}}!')
+      .withInput(undefined)
+      .toCompileTo('Goodbye\n\n!');
 
-    shouldCompileTo(
-      '{{#unless foo}}Goodbye{{../test}}{{test2}}{{/unless}}',
-      undefined,
-      'Goodbye'
-    );
+    expectTemplate('{{#unless foo}}Goodbye{{../test}}{{test2}}{{/unless}}')
+      .withInput(undefined)
+      .toCompileTo('Goodbye');
   });
 
   it('comments', function() {
-    shouldCompileTo(
-      '{{! Goodbye}}Goodbye\n{{cruel}}\n{{world}}!',
-      { cruel: 'cruel', world: 'world' },
-      'Goodbye\ncruel\nworld!',
-      'comments are ignored'
-    );
+    expectTemplate('{{! Goodbye}}Goodbye\n{{cruel}}\n{{world}}!')
+      .withInput({
+        cruel: 'cruel',
+        world: 'world'
+      })
+      .withMessage('comments are ignored')
+      .toCompileTo('Goodbye\ncruel\nworld!');
 
-    shouldCompileTo('    {{~! comment ~}}      blah', {}, 'blah');
-    shouldCompileTo('    {{~!-- long-comment --~}}      blah', {}, 'blah');
-    shouldCompileTo('    {{! comment ~}}      blah', {}, '    blah');
-    shouldCompileTo('    {{!-- long-comment --~}}      blah', {}, '    blah');
-    shouldCompileTo('    {{~! comment}}      blah', {}, '      blah');
-    shouldCompileTo('    {{~!-- long-comment --}}      blah', {}, '      blah');
+    expectTemplate('    {{~! comment ~}}      blah').toCompileTo('blah');
+    expectTemplate('    {{~!-- long-comment --~}}      blah').toCompileTo(
+      'blah'
+    );
+    expectTemplate('    {{! comment ~}}      blah').toCompileTo('    blah');
+    expectTemplate('    {{!-- long-comment --~}}      blah').toCompileTo(
+      '    blah'
+    );
+    expectTemplate('    {{~! comment}}      blah').toCompileTo('      blah');
+    expectTemplate('    {{~!-- long-comment --}}      blah').toCompileTo(
+      '      blah'
+    );
   });
 
   it('boolean', function() {
     var string = '{{#goodbye}}GOODBYE {{/goodbye}}cruel {{world}}!';
-    shouldCompileTo(
-      string,
-      { goodbye: true, world: 'world' },
-      'GOODBYE cruel world!',
-      'booleans show the contents when true'
-    );
+    expectTemplate(string)
+      .withInput({
+        goodbye: true,
+        world: 'world'
+      })
+      .withMessage('booleans show the contents when true')
+      .toCompileTo('GOODBYE cruel world!');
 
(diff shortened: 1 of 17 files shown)


jbboehr:
Looks like `HandlebarsTestBench` as-is ignores it's `message` property. Maybe this?

```diff
diff --git a/spec/env/common.js b/spec/env/common.js
index 5997322..1a78619 100644
--- a/spec/env/common.js
+++ b/spec/env/common.js
@@ -183,7 +183,7 @@ HandlebarsTestBench.prototype.withMessage = function(message) {
 };
 
 HandlebarsTestBench.prototype.toCompileTo = function(expectedOutputAsString) {
-  expect(this._compileAndExecute()).to.equal(expectedOutputAsString);
+  expect(this._compileAndExecute()).to.equal(expectedOutputAsString, this.message);
 };
 
 // see chai "to.throw" (https://www.chaijs.com/api/bdd/#method_throw)
```

nknapp:
First of all, I wouldn't say that this change has little benefit. I have often looked at test cases and had a hard time to understand what was actually tested, especially with the `shouldCompileTo`-method. 
That was the reason for me to introduce the `expectTemplate`-method in the first place.

I think replacing all calls (also to `shouldCompileTo`) is a valuable task and I am glad that you volunteer to do it. I thought I even had created an issue for it, but I cannot find it anymore...

Of course, the thing with tests is, that if you want to ensure backwards compatibility, you shouldn't change them. But on the other hand, it will make the code much more readable. The important thing is that they are really semantically equal to the old version.

I haven't had a detailed look at your PR yet, but please proceed with replacing `shouldCompileTo` as well.

nknapp:
> Looks like `HandlebarsTestBench` as-is ignores it's `message` property. Maybe this?
> 
> ```diff
> diff --git a/spec/env/common.js b/spec/env/common.js
> index 5997322..1a78619 100644
> --- a/spec/env/common.js
> +++ b/spec/env/common.js
> @@ -183,7 +183,7 @@ HandlebarsTestBench.prototype.withMessage = function(message) {
>  };
>  
>  HandlebarsTestBench.prototype.toCompileTo = function(expectedOutputAsString) {
> -  expect(this._compileAndExecute()).to.equal(expectedOutputAsString);
> +  expect(this._compileAndExecute()).to.equal(expectedOutputAsString, this.message);
>  };
>  
>  // see chai "to.throw" (https://www.chaijs.com/api/bdd/#method_throw)
> ```

Yes, that's right. I don't know why I didn't use it. Maybe do this as a separate commit so that it can be more easily reverted. I would even discard the third parameter of `toThrow` in favor of using `this.message`

nknapp:
I should probably also share my vision about the spec and about language test-cases:

My long-term goal which also addresses #1277, is to write a specification text, but also include test-cases in yaml form (independent of any language). An example is [this file](https://github.com/handlebars-lang/specification/blob/master/sections/03-01-basic-context.yaml), but this is just an early thought. It would be cool to have some kind of IDE support to write those examples, but not strictly necessary.

Helpers are, of course, difficult to describe that way, but this can probably be solved similar to the [lambdas in the mustache-spec](https://github.com/mustache/spec/blob/master/specs/~lambdas.yml).

I want to mention this, because rewriting all tests using `expectTemplate`, and your goal of generating JSON, is a first step into this direction and most of the tests should somehow be part of the spec.

I should probably start setting up another VuePress site containing the language spec.

jbboehr:
> First of all, I wouldn't say that this change has little benefit.

Well I just felt a little bad waltzing in here and changing the git-blame on the entire test suite, that's all :)

> Of course, the thing with tests is, that if you want to ensure backwards compatibility, you shouldn't change them.

Should be pretty safe. I could try to write some code that verified the inputs to CompilerContext were binary-equal, but that seems like overkill.

> Maybe do this as a separate commit so that it can be more easily reverted. I would even discard the third parameter of toThrow in favor of using this.message

Done.

> My long-term goal [...] is to write a specification text, but also include test-cases in yaml form (independent of any language)

That would be great! I'll have to stick to my Frankenstein creation in [[0]] in the meantime, of course.

I hope you'll also throw up a page with a list of all the different implementations at some point, if there isn't already. They can be hard to track down sometimes.

> Helpers are, of course, difficult to describe that way, but this can probably be solved similar to the lambdas in the mustache-spec.

Feel free to ping me when you need a bunch of javascript converted to PHP. For C, I have to resort to this abomination [[2]].


---


A quick overview of my recent commits. In the interest of avoiding transcription errors, I wrote a converter [[1]] to convert `shouldCompileTo` and `shouldCompileToWithPartials` to use `expectTemplate`.

There are currently individual commits for:
- the original large batch of `equals` and `shouldThrow` I ported by hand to use `expectTemplate` (wasn't really possible to do any automatic conversion)
- the discussed changes to `message` in the test bench
- two different test cases my converter had trouble with I ported by hand
- adding `withDecorator` and `withDecorators` to the test bench
- one big commit with the automatic conversion

Of course, feel free to request any changes, rebases, rewords, squashes, merges, etc. I'm going to try and do another manual review of the whole thing, but it'll probably take a while.

[0]: https://github.com/jbboehr/handlebars-spec/
[1]: https://gist.github.com/jbboehr/00b0c336ba81483712706d0db6a13f09
[2]: https://github.com/jbboehr/handlebars.c/blob/e0e010fae1cdb9ccd43ae6e391d30cfc7668abe5/tests/fixtures.c#L1577-L1579

jbboehr:
@nknapp Sorry there wasn't any context, typing up my previous post took a while, I should have prepared it beforehand.

> Sadly also a lot of work to review

Yeah, I was a little worried doing most of the work in advance that the review burden would be too high to have it merged :S

jbboehr:
Perhaps off-topic, but going over the partial tests and there is this one test:

```javascript
  it('partials with undefined context', function() {
    var string = 'Dudes: {{>dude dudes}}';
    var partial = '{{foo}} Empty';
    var hash = {};
    shouldCompileToWithPartials(
      string,
      [hash, {}, { dude: partial }],
      true,
      'Dudes:  Empty'
    );
  });
```

but the context is `{}` not `undefined`.

spec/basic.js
```diff
@@ -6,117 +6,141 @@ beforeEach(function() {
...
 
   it('comments', function() {
-    shouldCompileTo(
-      '{{! Goodbye}}Goodbye\n{{cruel}}\n{{world}}!',
-      { cruel: 'cruel', world: 'world' },
-      'Goodbye\ncruel\nworld!',
-      'comments are ignored'
-    );
+    expectTemplate('{{! Goodbye}}Goodbye\n{{cruel}}\n{{world}}!')
+      .withInput({
```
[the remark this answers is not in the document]
jbboehr: Pretty sure - I'll double-check. But I never would have formatted this the way it is:

```javascript
it('should allow the #each-helper to be used along with partial-blocks', function() {
  expectTemplate(
    '<template>{{#> list value}}value = {{.}}{{/list}}</template>'
  )
    .withInput({
      value: ['a', 'b', 'c']
    })
    .withPartials({
      list:
        '<list>{{#each .}}<item>{{> @partial-block}}</item>{{/each}}</list>'
    })
    .toCompileTo(
      '<template><list><item>value = a</item><item>value = b</item><item>value = c</item></list></template>'
    );
});
```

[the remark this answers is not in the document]
jbboehr: Actually, the one you referenced looks pretty good to me. What did you expect it to look like?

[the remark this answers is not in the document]
nknapp: It changed from 
```
{ cruel: 'cruel', world: 'world' }
```
to 
```
{ 
  cruel: 'cruel', 
  world: 'world'
},

and I thought it shouldn't. But maybe the line just got a little longer and crossed the threshold.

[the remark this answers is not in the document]
Comment (nknapp) [reply spec/basic.js:64]: I'll check this later. Don't worry about it.

**Comment by nknapp** [comment]
Apart from the review, I have been wondering about these lines: https://github.com/handlebars-lang/handlebars.js/blob/master/spec/builtins.js#L228-L243

It would be really cool to get completely remove "compileWithPartials", but this call cannot be replaced easier: I am thinking about adding a new method to `expectTemplate`:

* Either `.toCompileToAnyOf(expectedResult1, expectedResult2, ...)` 
* or something that can be called like `.toCompileToMatching(result => result === expectedResult1 || result === expectedResult2)`

Do you have any preference in this? The first one is more explicite, and since you are not using JavaScript, if would probably be more useful to you. The second one is more flexible, in case other use cases come along.

I think I would tend to use `.toCompileToAnyOf`

**Comment by jbboehr** [comment]
@nknapp I would lean towards the former. If you wanted something more generic, using a regex would allow a bit more flexibility while still being used easily from other languages.

**Comment by jbboehr** [comment]
Looks like the build failed due to a timeout in `tasks/task-tests/git.test.js`.

File: spec/data.js [hunk 2]
```diff
@@ -39,326 +32,253 @@ describe('data', function() {
...
-
     var partials = {
       myPartial: CompilerContext.compile('{{hello}}', { data: true })
     };
 
-    var helpers = {
-      hello: function(options) {
+    expectTemplate('{{>myPartial}}')
+      .withCompileOptions({ data: true })
+      .withPartials(partials)
```
nknapp, line 141: Would be nicer if this would read

```
.withPartial('myPartial' , '{{hello}}')
```
if that works

spec/data.js
```diff
@@ -39,326 +32,253 @@ describe('data', function() {
...
+      .withMessage('Overriden data output by helper')
+      .toCompileTo('sad world?');
   });
 
   describe('@root', function() {
     it('the root context can be looked up via @root', function() {
-      var template = CompilerContext.compile('{{@root.foo}}');
-      var result = template({ foo: 'hello' }, { data: {} });
-      equals('hello', result);
+      var string = '{{@root.foo}}';
```
nknapp, line 245: I would inline "string" and "input".

I think some code duplication is OK in tests, if it makes the test easier readable.

spec/data.js
```diff
@@ -39,326 +32,253 @@ describe('data', function() {
...
+        .toCompileTo('hello');
     });
+
     it('passed root values take priority', function() {
-      var template = CompilerContext.compile('{{@root.foo}}');
-      var result = template({}, { data: { root: { foo: 'hello' } } });
-      equals('hello', result);
+      expectTemplate('{{@root.foo}}')
+        .withRuntimeOptions({ data: { root: { foo: 'hello' } } })
+        .toCompileTo('hello');
```
nknapp, line 261: I tried to stick to the goal of this PR up to now, which is to convert tests to "expectTemplate" and not modify them otherwise. But reading this test-description and looking at the test, I think it is incomplete.

The input should be `{ foo: 'some other value' }`, in order to verify that the value from `data` is used, even if a root-value exists. Could you add that?

nknapp:
> @nknapp I would lean towards the former. If you wanted something more generic, using a regex would allow a bit more flexibility while still being used easily from other languages.

Good. I will see if I change it, but rather in a separate PR.

nknapp:
> Looks like the build failed due to a timeout in `tasks/task-tests/git.test.js`.

I'll see to that in the end, when I have reviewed all files. Maybe it helps to just increase the timeout.

spec/data.js
jbboehr: We could also do, possibly, with some small changes:

```javascript
      expectTemplate(string)
        .withInput(input)
        .withRuntimeOptions({ data: {} })
        .toCompileTo('hello')

        .withRuntimeOptions({})
        .toCompileTo('hello');
```

not sure it's more readable but it reduces duplication...

spec/data.js
nknapp: :+1:

Issue #1686, opened 2020-05-05
Nested each with equal values skips some iterations.


graphicore:
Before filing issues, please check the following points first:

- [ x] Please don't open issues for security issues. Instead, file a report at https://www.npmjs.com/advisories/report?package=handlebars
- [x ] Have a look at https://github.com/wycats/handlebars.js/blob/master/CONTRIBUTING.md
- [x ] Read the FAQ at https://github.com/wycats/handlebars.js/blob/master/FAQ.md
- [x ] Use the jsfiddle-template at https://jsfiddle.net/4nbwjaqz/4/ to reproduce problems or bugs

This will probably help you to get a solution faster. 
For bugs, it would be great to have a PR with a failing test-case.

Here's the jsfiddle for the issue: https://jsfiddle.net/graphicore/ge1qxvc8/

### Template

```
_{{#each abc as | c1 |}}{{#each ../abc as | c2 |}}{{#each ../../abc as | c3 |}}{{c1}}{{c2}}{{c3}} {{else}}! {{/each}}| {{/each}}{{/each}}_
```

### Data

```
{
    abc: ['A', 'B', 'C']
}
```

### Observed Behavior

```
_! | ABA ABB ABC | ACA ACB ACC | BAA BAB BAC | ! | BCA BCB BCC | CAA CAB CAC | CBA CBB CBC | ! | _ 
```

### Expected Behavior

```
_AAA AAB AAC | ABA ABB ABC | ACA ACB ACC | BAA BAB BAC | BBA BBB BBC | BCA BCB BCC | CAA CAB CAC | CBA CBB CBC | CCA CCB CCC | _ 
```

### Context:

I'm porting a [font testing tool](https://github.com/impallari/Font-Testing-Page) written mostly in server side PHP to [client side JavaScript](https://github.com/graphicore/specimenTools/pull/35). For some samples I need to generate grids of char combinations e.g. to let a designer test spacing [uc-spacing-03.php](https://github.com/impallari/Font-Testing-Page/blob/master/includes/latin-02/uc-spacing-03.php).

I believe 279e038ba7ff7e5966a60e36d326e2d4c310f0c6 is related (commit message).

graphicore:
I  added a test and suggested a fix in PR #1687

nknapp:
Is this a duplicate of #1300 ?

graphicore:
@nknapp I looked at that issue before, but it's just now that I've noticed your workaround, and indeed it works.

[...]

PR #1687, opened 2020-05-05
Add failing unit test for #1686

graphicore:

- [ ] Maintain the existing code style
- [ x] Are focused on a single change (i.e. avoid large refactoring or style adjustments in untouched code if not the primary goal of the pull request)
- [ x] Have tests
- [ x] Don't significantly decrease the current code coverage (see coverage/lcov-report/index.html)
- [ x] Currently, the `4.x`-branch contains the latest version. Please target that branch in the PR.


graphicore:
Note, the proposed fix in 2981de4 is just a suggestion. I'm really not deep enough into the code base to judge if the proposed fix is doing any harm. I put an explanation into the commit message:

>  Add explicit goDeeper option to wraped program;Fixes #1686
>
>According to the commit message of 279e038:
>
>" [...] Helper authors now need to take care to return the same context
>value whenever it is conceptually the same and to avoid behaviors that
>may execute children under the current context in some situations and
>under different contexts under other situations."
>
>I belive that metric is too implicit, as the issue demonstrates. This
>is because in this case the context value is not conceptually the same,
>but it is de facto the same value.
>
>Instead this suggests to use an explicit option flag "goDeeper" to mark
>when a helper definitely should go deeper. This also should be backwards
>safe, as it doesn't change the behavior of existing helpers, builtin or
>custom, except the builtin "each",

Issue #1593, opened 2019-11-04
global is not defined in each.js


dennisblack666:
After the version 4.1.2 there is an error in the each.js helper file. 

> **ReferenceError: global is not defined**

![image](https://user-images.githubusercontent.com/37075205/68115083-d7a80200-fef7-11e9-9bee-008f468878d4.png)

nknapp:
That's strange because there was no problem with the build and the tests. 

Who is showing this error measure? You IDE? Eslint?

dennisblack666:
The bug is in the Google Chrome Console when the context is an object.

When i remove the "global." property it works correct (jumps into the else statement) and the template shows the correct values, without the ReferenceError.

nknapp:
Could you describe exactly, what you are doing? Which file are you looking at (which URL)? In the built handlebar.js file, the each helper is wrapped with a function by Webpack, which defines "global" as a parameter". I cannot reproduce your issue anywhere.

dennisblack666:
Found the issue in our webpack.config. Changed the property "node: false" to "node: {global: true}".

Thank you for your help!

kjg:
I'm having this same issue. The specs define `global`
https://github.com/handlebars-lang/handlebars.js/blob/0c8230c2539b6815e318b5f2b85666a49baaefc6/spec/env/common.js#L1
 which is why this doesn't cause any issues in the tests.

Is it expected that we should have to always define `global` if we use handlebars?

nknapp:
OK, the change was made when #1557 was merged. So, this is a regression. As you say, there wasn't a test to catch this problem and I hadn't seen this coming when I merged the PR.

If you have a suggestion on how to solve this and you are willing to open a PR, go ahead.

I think there should be 3 steps in a PR:

1. **Have a test**. Since the setup for unit-tests is pretty fixed, and this issue depends on webpack configuration, it should be an integration-test. The folder [integration-testing](https://github.com/handlebars-lang/handlebars.js/tree/4.x/integration-testing) already contains some tests that test handlebars with different webpack and babel configs. Maybe you could add a new one that reproduces this problem.

2. **Implement a solution**. Do you have an idea how to solve this? I believe it could be done by checking `typeof Symbol` instead of `global.Symbol`, but I am not sure.

3. **Add an eslint-rule**: I don't know if there is an eslint-plugin that would prevent such an error in the future. But if there is, we should integrate it.

kjg:
[...]

PR #1688, opened 2020-05-07
refactor: do not create a "global" variable in tests

nknapp:
see #1593


jaylinski:
Closing because the original issue is already fixed.

PR #1743, opened 2021-04-02
Fix Grammar in "compiler-api"

s-mrb:
What was the issue?

Following Line have grammar error (``give`` instead of`` given``)

``
Used to generate the code to resolve a give path component.
``


jaylinski:
Thanks!

Issue #1736, opened 2021-02-13
RCE possible in compat and strict mode


nknapp:
Two security issues have arised and are fixed in the referencing commits:

1) Due to insufficient escaping of the input template, it was possible to inject code into templates that are compiled in "compat" mode.

2) In "strict" mode, the exploits disclosed in the npm-security advisories [755](https://www.npmjs.com/advisories/755),
[1164](https://www.npmjs.com/advisories/1164), [1316](https://www.npmjs.com/advisories/1316),
[1324](https://www.npmjs.com/advisories/1324) and [1325](https://www.npmjs.com/advisories/1325) and in the blog-article
of [Mahmoud Gamal](http://mahmoudsec.blogspot.com/2019/04/handlebars-template-injection-and-rce.html) possible, because the the method that was used in strict-mode had not called the safe-guard methods.

The issues have been disclosed a couple of weeks ago at https://snyk.io/vuln/SNYK-JS-HANDLEBARS-1056767 and are fixed in version 4.7.7

joshbressers:
This issue is getting flagged by security scanners (the commits have scary words in them it would seem). Can you clarify if these are indeed security fixes?

Thanks in advance

nknapp:
Yes, these are indeed security fixes, although only relevant in "compat" and "strict" mode.
 I'm not sure if they are already disclosed. My plan is to fill this issue with more details and links when they are.

jtnord:
is there any plan to backport any relevant fixes to the 3.x branch, or is that line considered dead now?

jtnord:
@nknapp seems like 3.0.8 should have these fixes already due to #1532  ?  can you confirm this is the case and that the [CVE cpe metadata](https://nvd.nist.gov/vuln/detail/CVE-2021-23369/cpes?expandCpeRanges=true) is incorrect?

bitwiseman:
I've attempted a backport of these fixes - #1751 .

jaylinski:
Note: v3.x is EOL and did not receive all relevant security fixes. Upgrade to at least v4.7.7.

PR #1751, opened 2021-05-19
Backport security fixes for #1736

bitwiseman:
I would not file a PR for this except that the security advisory has already been filed. 
This PR addresses the vulnerability in `compat` mode and documents the lack of vulnerability in `strict` mode. 
@nknapp 

- [ ] Please don't start pull requests for security issues. Instead, file a report at https://www.npmjs.com/advisories/report?package=handlebars
- [x] Have tests
- [ ] Currently, the `4.x`-branch contains the latest version. Please target that branch in the PR.


spec/security.js
```diff
@@ -24,11 +37,55 @@ describe('security issues', function() {
...
+        it('should not allow constructors to be accessed', function() {
+          expectTemplate('{{lookup (lookup this "constructor") "name"}}')
+            .withCompileOptions(compileOptions)
+            .withInput({})
+            .toCompileTo('');
+          if (compileOptions.strict) {
+            expectTemplate('{{constructor.name}}')
+              .withCompileOptions(compileOptions)
+              .withInput({})
+              .toThrow(TypeError);
```
bitwiseman, line 61: This was already the case in 3.x before this change and remains after this change currently.   `strict` mode resulted in a `TypeError` when accessing `constructor`, `__proto__`, etc.  We can fix this but it would require more changes.

lib/handlebars/runtime.js
```diff
@@ -69,12 +69,28 @@ export function template(templateSpec, env) {
...
+        return result;
+      }
+
+      return undefined;
     },
     lookup: function(depths, name) {
       const len = depths.length;
       for (let i = 0; i < len; i++) {
-        if (depths[i] && depths[i][name] != null) {
+        let result = depths[i] && container.lookupProperty(depths[i], name);
```
bitwiseman, line 92: This covers compat mode.

lib/handlebars/runtime.js
```diff
@@ -69,12 +69,28 @@ export function template(templateSpec, env) {
       if (!(name in obj)) {
         throw new Exception('"' + name + '" not defined in ' + obj);
       }
-      return obj[name];
+      return container.lookupProperty(obj, name);
```
bitwiseman, line 72: It appears `strict` mode is already covered by the check in `javascript-compiler.js`:

https://github.com/handlebars-lang/handlebars.js/blob/16bd606fec0ada4c811eb7afe6bfd619d88f3d37/lib/handlebars/compiler/javascript-compiler.js#L16-L19

However, the reason why this vulnerability appeared in 4.x was because that check was removed and this change was present. Better to make this change to make it completely clear.

batmat:
@nknapp app hello! Any chance you would be able to look into this some day soon? We are ready to help as need be, please let us know :-).
Thanks a lot for your hard work!

bitwiseman:
@ErisDS @Afanyiyu @kpdecker 
Sorry for pinging you but this is a security issue that it would be great to get backported sooner rather than later.

bitwiseman:
@jaylinski 
Hey, thanks for the label update.  If there is anything I can do to make this PR merge sooner, please just say so.  
@car-roll @batmat @jtnord may also want to be updated on this.

jaylinski:
Memo to myself:

* [ ] Think about back-porting GitHub CI code for tests
* [ ] Don't use npm lock-file version v2, keep it to v1
* [ ] Test fixes against exploits (https://snyk.io/vuln/npm:handlebars)

nknapp:
@bitwiseman I think the compat-mode (f0589701698268578199be25285b2ebea1c1e427) part is worth backporting, but I wouldn't touch the part that is no real security issue in version 3.

Changes in version 3 are only done on a strict "need to be there" basis, as far as I am concerned...

nknapp:
can you modify your PR, so that only  f058970 is backported?

windusayles:
@bitwiseman @nknapp do we need the modification you mentioned, f058970, in order to merge this fix?

wycats:
@windusayles yeah 👍 

Can you update the PR to limit it to just https://github.com/handlebars-lang/handlebars.js/commit/f0589701698268578199be25285b2ebea1c1e427?

jaylinski:
Closing since v3.0.8 only has a few thousand downloads a week. So most people moved on to the latest (and more secure) version.

image

https://www.npmjs.com/package/handlebars?activeTab=versions

Also: v3.x is now marked as EOL in our security policy (35f001862c1667af978347d78c0b8a6115b01766).

PR #1752, opened 2021-05-21
runtime: use custom escape function

Cireo:
Without this change, the line

    hb.escapeExpression = Utils.escapeExpression;

in top-level `create()` is unused.

Users who use a custom escape can only do so by modifying the global Utils.escapeExpression.


Cireo:
Wrong branch, started on master

PR #1753, opened 2021-05-25
runtime: use custom escape function

Cireo:
Without this change, the line

    hb.escapeExpression = Utils.escapeExpression;

in top-level `create()` is unused.

Users who use a custom escape can only do so by modifying/corrupting
the global Utils.escapeExpression, or using deep copies.

### DIFF
--- a/lib/handlebars/runtime.js
+++ b/lib/handlebars/runtime.js
@@ -153,7 +153,7 @@ export function template(templateSpec, env) {
       return typeof current === 'function' ? current.call(context) : current;
     },
 
-    escapeExpression: Utils.escapeExpression,
+    escapeExpression: env.escapeExpression,
     invokePartial: invokePartialWrapper,
 
     fn: function(i) {


Cireo:
@jaylinski is there something I should have done to request a review of this?

jaylinski:
@Cireo It's okay, but we have to take a closer look at it.

Adding an automated test for this is probably hard, right?

Cireo:
No, it only takes a few lines to demonstrate the existing bug.

This doesn't add any new functionality, it repairs a slightly broken feature that was never properly implemented.
``` node
> const clobberEscape = (val) => "escaped"
> const test = (hb) => hb.compile("x {{y}} z")({y: "y"})

> const Handlebars = require('handlebars');

// Pass in escapeExpression
> const hb1 = Handlebars.create({escapeExpression: clobberEscape});
> test(hb1)
'x y z'  // BUG: expect "x escaped z"

> const hb2 = Handlebars.create()  // workaround by modifying global escape
> hb2.Utils.escapeExpression = clobberEscape;
> test(hb2)
'x escaped z'
```

Cireo:
It's been months since I did the initial debugging, so I don't recall if you're supposed to be able to pass in the `escapeExpression` to `create` or what the exact syntax is.

jaylinski:
Duplicate of #1523.

PR #1793, opened 2021-12-04
Replace Saucelabs with Playwright

jaylinski:
Also reorganized test-folders.


tests/server.js
```diff
@@ -0,0 +1,36 @@
+const fs = require('fs');
```
nknapp, line 1: Do we need a custom written server here? Couldn't we just use "grunt connect" instead? Or something like "serve" or "sirv-cli"

.gitignore
```diff
@@ -16,5 +15,5 @@ node_modules
 lib/handlebars/compiler/parser.js
 /coverage/
 /dist/
-/integration-testing/*/dist/
+/tests/integration/*/dist/
 /spec/tmp/*
```
nknapp, line 19: Should have a newline.

nknapp (commented):
There is a lot of cleanup in this one commit, along with the actual changes. That makes it hard to track the actual changes.

It would be easier to review if there were separate commits in thie PR: One for "renaming and moving stuff", one for "replacing saucelabs with playwright".

Not in this PR anymore...

I left two remarks. Please think about them. Other than that I think the changes are good and the renaming and moving files is beneficial for understanding the project.

We should try to move away from grunt...

.gitignore
jaylinski: Done.

tests/server.js
jaylinski: I also thought about this. I created a custom server because, as you said

> We should try to move away from grunt...

and because I'm not a fan big dependency trees. `sirv` seems nice though, if you prefer it, I'll replace the custom solution.

jaylinski:
> It would be easier to review if there were separate commits in thie PR

@nknapp I split the changes in multiple commits, if you want to take another look.

Gruntfile.js
```diff
@@ -238,59 +196,42 @@ module.exports = function(grunt) {
...
-  } else {
-    grunt.registerTask('sauce', []);
-  }
-
-  // Requires secret properties (saucelabs-credentials etc.) from .travis.yaml
+  // Requires secret properties from .travis.yaml
   grunt.registerTask('extensive-tests-and-publish-to-aws', [
     'default',
     'bgShell:integrationTests',
-    'sauce',
```
nknapp, line 280: This is ok for now, but in the future, we should make sure that the browser tests have run before publishing to aws.

Maybe we can model the whole deployment workflow in Github-Actions. I have created an issue for this, so that we don't forget. #1806

nknapp (approved):

nknapp (changes requested):
custom-server should be replaced with grunt-connect (see my comment there). We can replace it later, when we know better in which direction we are going

Gruntfile.js
nknapp: you don't have to change anything here. I just wanted to let you know about my thoughts.

jaylinski: > Using "connect" would just mean to call "grunt connect" instead of "tests/server.js", correct?

Yep.

jaylinski:
> custom-server should be replaced with grunt-connect (see my comment there). We can replace it later, when we know better in which direction we are going

@nknapp Done in 315c1ad05617a0608df1f9f165f27565d59f7ff9.

tests/server.js
jaylinski: Fixed with commit 315c1ad05617a0608df1f9f165f27565d59f7ff9.

PR #1805, opened 2021-12-23
The changes were applied to the precompiler before the file was compiled

ErfanKamalian:
### How to fix it

To solve this problem, the explanations related to it were read first.
After studying the code, searching, researching and developing (R&D) about this problem, it was finally solved as follows:
Regex documents were studied and designed string with existing examples that can find \n and applied using the replace function.
Then, to optimize and not implement the dirty code, this function is called on the final file before compilation in precompiler.js.

Erfan Kamalian
web programming course -- Sharif University of Technology

### DIFF
--- a/lib/precompiler.js
+++ b/lib/precompiler.js
@@ -22,6 +22,21 @@ module.exports.loadTemplates = function(opts, callback) {
   });
 };
 
+// Erfan Kamalian
+
+function myNewLine(string) {
+
+
+  // use regex format to replace all \n with <br> in clean way!  
+
+  const out = string.replace(/(?:\r\n|\r|\n)/g, '<br>');
+
+
+  // return output
+
+  return out;
+}
+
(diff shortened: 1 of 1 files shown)


jaylinski:
@ErfanKamalian Thank you for your contribution.

As stated in https://github.com/handlebars-lang/handlebars.js/issues/1738#issuecomment-1000545517, this can be achieved by using a custom helper.

Since it can be a big pain to add a custom helper to all template-variables, we are thinking about adding a new `hooks`-function to allow global customizations of variables (https://github.com/handlebars-lang/handlebars.js/pull/1523#issuecomment-548119984).

ErfanKamalian:
Thank you for your response.
I enjoyed this project and learned a lot. I agree with you on this problem as well.

I hope we can work together more. 
I have just started web programming :)

PR #1809, opened 2021-12-29
Enable IDE code-highlighting for bin-file

jaylinski:
By adding the `.js`-extension to the bin-file, IDEs will
know that js code-highlighting should be applied to the
file, which makes it easier for devs to read the code.

For comparison, other JS libraries also use the `.js`-extension:

* https://github.com/webpack/webpack/blob/v5.65.0/bin/webpack.js
* https://github.com/webpack/webpack/blob/v5.65.0/package.json#L126

### DIFF
--- a/spec/expected/help.menu.txt
+++ b/spec/expected/help.menu.txt
@@ -1,5 +1,5 @@
 Precompile handlebar templates.
-Usage: handlebars [template|directory]...
+Usage: handlebars.js [template|directory]...
 
 Options:
   --help               Outputs this message                                                                    [boolean]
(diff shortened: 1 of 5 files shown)


spec/expected/help.menu.txt
```diff
@@ -1,5 +1,5 @@
 Precompile handlebar templates.
-Usage: handlebars [template|directory]...
+Usage: handlebars.js [template|directory]...
```
jaylinski, line 2: :warning: This shouldn't change.

jaylinski: This works as intended. From the `yargs` docs:

*.usage(, [desc], [builder], [handler])*
```
Set a usage message to show which commands to use. Inside message, the string $0 will get interpolated to the current script name or node command for the present script similar to how $0 works in bash or perl.
```

Since npm aliases the bin-script with a without a `.js`-extension, the output won't show `.js` if called via npm/npx.

PR #1813, opened 2022-01-01
Use latest `source-map`-version

### LINKED ISSUE #1427: Please upgrade source-map to 0.7.2 to get speed improvements
The latest source-map which is v0.7.2 got support for WebAssembly and it is 10 times faster than the old one. More info here:

https://github.com/mozilla/source-map/blob/master/CHANGELOG.md#072
http://fitzgeraldnick.com/2018/02/26/speed-without-wizardry.html
https://hacks.mozilla.org/2018/01/oxidizing-source-maps-with-rust-and-webassembly/
https://github.com/devtools-html/debugger.html/issues/5598

Please update so you would get 10 times speed improvements for your library in Firefox, 9 times in Safari and 5 times in Chrome.

jaylinski:
Resolves #1427.


jaylinski:
Waiting until build script was updated, because the current babel-webpack-setup can't handle `async/await`.

PR #1815, opened 2022-01-02
Remove support for IE and dead browsers, get rid of `underscore`-dependency in specs, upgrade to prettier v2 and support worker-envs

jaylinski:
Handlebars v5 will target modern browsers,
while older browsers will still receive support
via version 4.x.


README.markdown
```diff
@@ -79,17 +79,17 @@ There are a few Mustache behaviors that Handlebars does not implement.
 Supported Environments
 ----------------------
 
-Handlebars has been designed to work in any ECMAScript 3 environment. This includes
+Handlebars has been designed to work in any ECMAScript 7 (2016) environment. This includes
 
-- Node.js 10+
+- Node.js
+- Deno
```
jaylinski, line 85: Add a test for this!

jaylinski: Too complicated, won't support.

PR #1816, opened 2022-01-03
[WIP] Add possibility to precompile as ES module

jaylinski:
### TODO

* [ ] Write tests

### DIFF
--- a/bin/handlebars.js
+++ b/bin/handlebars.js
@@ -11,10 +11,11 @@ const yargs = require('yargs')
     type: 'string',
     description: 'Source Map File'
   })
-  .option('a', {
-    type: 'boolean',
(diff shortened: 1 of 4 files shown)


Sparticuz:
Is there anything else that needs to be done to support esm precompiled templates? Just switching out the requires to an import doesn't seem to work. I'm guessing there should be some type of export in the file?

jaylinski:
It seems you answered your question yourself in https://github.com/handlebars-lang/handlebars.js/issues/1723#issuecomment-1531764033. I'll incorporate it in this PR (and hopefully get it merged soonish).

zach-betz-hln:
Looking forward to this 🙂

PR #1864, opened 2022-05-17
Test on Node 18 and drop EOL Node 10

jaylinski:
:octocat:

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   "license": "MIT",
   "readmeFilename": "README.markdown",
   "engines": {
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
   "license": "MIT",
```
jaylinski, line 17: I'm not sure why yet, but the update to `@playwright/test` >= 1.20 causes an issue with UMD browser-tests.

jaylinski: Problem fixed itself with `@playwright/test` 1.27.

PR #1866, opened 2022-05-20
Use `@handlebars/parser` v2

jaylinski:
See https://github.com/handlebars-lang/handlebars-parser/releases/tag/v2.0.0


jaylinski:
Seem the ts compilation (https://github.com/handlebars-lang/handlebars-parser/blob/master/package.json#L23) does something that rollup doesn't like.

Issue #1840, opened 2022-03-21
Using key / context in helpers


btalb:
Consider the following two uses of a custom helper that needs to know information about the evaluation context in which it's been run:
```
{{#each robots}}
{{helperThatNeedsToKnowWhichRobot this}}
{{/each}}
```
```
{{#with robots.husky}}
{{helperThatNeedsToKnowWhichRobot this}}
{{/with}}
```
In the first case, my helper receives an options object with the following useful data fields:
```
    data: {                                                                                              
      root: [Object],                                                                                    
      _parent: [Object],                            
      key: 'husky',                                 
      index: 2,                                     
      first: false,                                 
      last: false                                   
    },  
```

But the second example gives me nothing to work with:
```
    data: { root: [Object] }
```

How can I implement the described helper which will also work in a `with` evaluation context?

jaylinski:
If I understand you correctly, you want to access the `robots`-object inside the `robots.husky` as a `_parent`-object when using the `with`-helper?

I guess that's a reasonable feature request. Should be doable by simply copying these lines to the `with.js`-helper:
https://github.com/handlebars-lang/handlebars.js/blob/v4.7.7/lib/handlebars/helpers/each.js#L32-L34

Would be nice if you could open a PR with a fix and a test-case. I'm not sure if this would be a breaking change, since some helpers may rely on the `_parent`-object not to be present. :thinking:

PR #1952, opened 2023-06-23
added _parent and context to data returned by #with in helpers #1840

smmhatami:
This is a fix to solve Issue 1840. Resolves https://github.com/handlebars-lang/handlebars.js/issues/1840. Added _parent and context to data returned by with in helpers just like each does to each iterated field. 
I just couldn't find where and how should I add tests for this functionality, i would be glad if you help me with that. 
thanks!

### DIFF
--- a/lib/handlebars/helpers/with.js
+++ b/lib/handlebars/helpers/with.js
@@ -1,5 +1,5 @@
 import { Exception } from '@handlebars/parser';
-import { isEmpty, isFunction } from '../utils';
+import { createFrame, isEmpty, isFunction } from '../utils';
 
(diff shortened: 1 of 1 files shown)


jaylinski (commented):
I think the right place for a test would be here:

https://github.com/handlebars-lang/handlebars.js/blob/b4a54ad86ed73130ac9a7663d2fb66accb360736/spec/builtins.js#L136-L184

Tests can be run via `npm install` and `npm run test`.

jaylinski (commented):
@smmhatami Can you add a test for this? (See comment above on where to add it.)

PR #1957, opened 2023-07-16
Fix #1748 where allowed prototype methods are not called

### LINKED ISSUE #1748: How to call String.prototype.trim (or any other prototype function) as shown in the documentation example?

I have been trying the example from the docs calling `String.prototype.trim` to work, but it does not seem to return the desired result (https://handlebarsjs.com/api-reference/runtime-options.html#options-to-control-prototype-access).

https://jsfiddle.net/3yahoj1d/

Can anyone explain why the result prints as `[object Object]` to the console when `typeof result` returns `string`, or how to get the example to work?

I have also asked this on StackOverflow (https://stackoverflow.com/q/67416982/5115898) so feel free to close this issue if this is not a problem with handlebars itself.

aalimovs:
Fixes #1748 where allowed prototype methods are not called.

Without this fix [the official documentation is incorrect](https://handlebarsjs.com/api-reference/runtime-options.html#options-to-control-prototype-access):

```js
const template = handlebars.compile("{{aString.trim}}");
const result = template(
  { aString: "  abc  " },
  {
    allowedProtoMethods: {
      trim: true
    }
  }
);
// result = 'abc'
```

The result is `[object Object]`, not `abc`.

### DIFF
--- a/lib/handlebars/runtime.js
+++ b/lib/handlebars/runtime.js
@@ -133,6 +133,9 @@ export function template(templateSpec, env) {
       }
 
       if (resultIsAllowed(result, container.protoAccessControl, propertyName)) {
+        if (typeof result === 'function') {
+          return parent[propertyName]();
+        }
         return result;
       }
       return undefined;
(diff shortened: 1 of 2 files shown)


aalimovs:
Branched out from wrong branch, will target `4.x`

PR #1958, opened 2023-07-16
Fix #1748 where allowed prototype methods are not called

### LINKED ISSUE #1748: How to call String.prototype.trim (or any other prototype function) as shown in the documentation example?

aalimovs:
Fixes #1748 where allowed prototype methods are not called.

Without this fix [the official documentation is incorrect](https://handlebarsjs.com/api-reference/runtime-options.html#options-to-control-prototype-access):

```js
const template = handlebars.compile("{{aString.trim}}");
const result = template(
  { aString: "  abc  " },
  {
    allowedProtoMethods: {
      trim: true
    }
  }
);
// result = 'abc'
```

The result is `[object Object]`, not `abc`.


aalimovs:
@jaylinski rebased, but I've no idea why this PR now picks up changes in `.github/workflows/ci.yml` that are already in `4.x` branch. I've tried rebasing the fork few different ways but still the same.

If you want me to recreate the PR let me know.

jaylinski:
No problem, I fixed it.

lib/handlebars/runtime.js
```diff
@@ -136,6 +136,9 @@ export function template(templateSpec, env) {
       }
 
       if (resultIsAllowed(result, container.protoAccessControl, propertyName)) {
+        if (typeof result === 'function') {
+          return parent[propertyName]();
```
jaylinski, line 140: Why did you use `parent[propertyName]()` instead of `result()`? Any specific reason?

aalimovs: It doesn't work with `result()`, complains that:

```
TypeError: String.prototype.trim called on null or undefined
```

I'm not 100% sure why, but as I understand `result` loses `this` context.

mohd-akram:
This fix isn't right - `lookupProperty` is only for looking up properties, not calling them. Functions are called at https://github.com/handlebars-lang/handlebars.js/blob/f422bfdf3ed321c32c6eecfaa778de46275fb76d/lib/handlebars/runtime.js#L146-L148

They are called with the Handlebars context, not with the object preceding the dot as in regular JavaScript. Changing this would be a breaking change, and it's not clear for what benefit. You can create a `trim` helper which achieves the same thing.

jaylinski:
I'm still curious why the documentation states that it should work. :thinking:

mohd-akram:
That was added somewhat recently to the documentation with the security fixes. I think it was just a mistake.

nknapp:
I don't know why I added this example to the docs and did not notice that it does not work.
The whole "allowProto..." options were just added to provide a way to get back the old behavior.  If it didn't work in pre 4.6 versions (and it doesn't) , I wouldn't fix it. 

The Handlebars way to solve the issue is to register a "trim" helper and use "{{trim aString}}" instead.

jaylinski:
I'll keep this issue open until we corrected the wrong documentation and close it afterwards.

PR #1977, opened 2023-08-03
[WIP] Coerce `noEscape`-statements to a string

### LINKED ISSUE #1838: noEscape cause additions of integer values
Hello,

When I set the parameter "noEscape" to true, it seems that tokens are added instead of concatenated. So if I reference integers inside tokens, an addition will be realised if no separation character is present between tokens.

According to the documentation, "noEscape" is just a way to disable HTML characters escaping, and I think that a solution would be to use a string version of the variables referenced by the tokens instead of the variables values directly.

Here is an example of the bug with multiple cases: https://jsfiddle.net/mhd0k4ne/70/

Thanks.

jaylinski:
Fixes #1838.


lib/handlebars/compiler/javascript-compiler.js
```diff
@@ -426,7 +426,7 @@ JavaScriptCompiler.prototype = {
     if (this.isInline()) {
       this.replaceStack(current => [' != null ? ', current, ' : ""']);
 
-      this.pushSource(this.appendToBuffer(this.popStack()));
+      this.pushSource(this.appendToBuffer([this.popStack(), '+', "''"]));
     } else {
       let local = this.popStack();
       this.pushSource([
```
jaylinski, line 432: What about non-inline statements (see below)? Do we also have to cast those to a string?

lib/handlebars/compiler/javascript-compiler.js
```diff
@@ -426,7 +426,7 @@ JavaScriptCompiler.prototype = {
     if (this.isInline()) {
       this.replaceStack(current => [' != null ? ', current, ' : ""']);
 
-      this.pushSource(this.appendToBuffer(this.popStack()));
+      this.pushSource(this.appendToBuffer([this.popStack(), '+', "''"]));
```
jaylinski, line 429: This changes the output of compiled templates, so tests have to be updated. Or a solution is found without altering templates.

nknapp: I don't see a way to do this without template changes. But we also don't need this in all places where "append" is called.
I believe BlockStatements and  PartialStatements are not affected because they always return strings.

So your solution looks simple, but I could probably combine `'+', "''"` into `"+''"`

Are the other tests passing? Seems like that they are not even run.

jaylinski: Thanks for the input!

There is the `test:bin` tests that check if the compiled templates match the expected compilation-output. I have to adapt those tests.

lib/handlebars/compiler/javascript-compiler.js
nknapp: Maybe it is better to add the empty string to "appendToBuffer", even though... I'll have another look. Something occurred to me

nknapp: I'll add discussion to the ticket. I think there might be a better solution, but I haven't grasped it completely yet

jaylinski: I sided with your solution and pushed it.

lib/handlebars/compiler/javascript-compiler.js
jaylinski: I fixed the `test:bin` tests.

PR #1966, opened 2023-07-27
Release v4.7.8

jaylinski:
It is time...

for a new release.

This PR prepares a new patch-release and updates release-instructions.

Note: the release system will be kept as is for the `4.x`-branch but will be updated for `5.x`.

Closes #1918.

### Commit diff

https://github.com/handlebars-lang/handlebars.js/compare/v4.7.7...4.x

### npm publish diff

`npm diff --diff=handlebars@4.7.7 --diff=handlebars-jaylinski@4.7.8`

### Comparison with previous release

https://github.com/handlebars-lang/handlebars.js/commit/a9a8e403213583ca90cb7c872d3a22796c37d961

PR #1855, opened 2022-05-04
fix: fix error for each helper with block params and strict mode #1820

arturmakaranka:
Fix for #each helper with params in strict mode. By the same pr of @jaylinski https://github.com/handlebars-lang/handlebars.js/pull/1821/files

PR #1979, opened 2023-08-04
refactor: rename i to startPartIndex

nknapp:
As said in #1966 I renamed "i" to "startPartIndex", which is maybe a little more helpful.
Minor additional changes it the use of "i" only within the for-loops and the use of "len" instead of "i" after the loop.

"i" and "len" should always be equal in this case.

### DIFF
--- a/lib/handlebars/compiler/javascript-compiler.js
+++ b/lib/handlebars/compiler/javascript-compiler.js
@@ -535,16 +535,22 @@ JavaScriptCompiler.prototype = {
     this.resolvePath('data', parts, 0, true, strict);
   },
 
-  resolvePath: function(type, parts, i, falsy, strict) {
+  resolvePath: function(type, parts, startPartIndex, falsy, strict) {
     if (this.options.strict || this.options.assumeObjects) {
       this.push(
-        strictLookup(this.options.strict && strict, this, parts, i, type)
+        strictLookup(
+          this.options.strict && strict,
+          this,
+          parts,
+          startPartIndex,
+          type
+        )
       );
       return;
     }
 
     let len = parts.length;
-    for (; i < len; i++) {
+    for (let i = startPartIndex; i < len; i++) {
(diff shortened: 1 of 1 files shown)


lib/handlebars/compiler/javascript-compiler.js
```diff
@@ -535,16 +535,22 @@ JavaScriptCompiler.prototype = {
...
+          startPartIndex,
+          type
+        )
       );
       return;
     }
 
     let len = parts.length;
-    for (; i < len; i++) {
+    for (let i = startPartIndex; i < len; i++) {
```
jaylinski, line 553: Why declare `i` here instead of using `parts[startPartIndex]` at line 556?

lib/handlebars/compiler/javascript-compiler.js
```diff
@@ -1280,7 +1286,7 @@ function strictLookup(requireTerminal, compiler, parts, i, type) {
       '(',
       stack,
       ', ',
-      compiler.quotedString(parts[i]),
+      compiler.quotedString(parts[len]),
```
jaylinski, line 1289: Can you explain this change to me? Before #1855, this was always `parts[0]` (because `i` was `0`). The fix changed this to be the start-part-index. Now you changed it to be `parts.length`.

Shouldn't a test have failed somewhere?

lib/handlebars/compiler/javascript-compiler.js
nknapp: I think it is better style not to use the parameter as variable, but instead us an ordinary for-loop with a pattern that everybody understands

lib/handlebars/compiler/javascript-compiler.js
nknapp: after the for-loop, after the `for (; i < len; i++) { ... }`, "i" always has the same value as "len", even before #1855.

PR #1980, opened 2023-08-04
Draft: Generate s3 file listing after each upload

nknapp:
# This PR is not ready yet. I am currently playing with the S3 api.

In order to fix the [builds page](https://com.s3.amazonaws.com/builds.handlebarsjs/bucket-listing.html?sort=lastmod&sortdir=desc) we should statically generate a file-index on each upload, so that we do not have to rely on the s3-api being available from the browser.

Since we are also relying on an old version of the aws-sdk, this will probably also include an update to the newest version.


nknapp:
@jaylinski if you like, you can have a look at this already.  If have extracted the s3-logic to grunt-independent functions and added an s3-listing function. With this, we should be able to generate some kind of listing-file, be it json, or, using Handlebars, directly a static html.

nknapp:
@jaylinski I have manually generated and deployed a static file-list to the s3-bucket using the code from this PR. This means we have a file-list now, but it will be outdated on the next build.

I have also fixed the link from the docs to the file-list.

The HTML file-list is generated via Handlebars, but the template and the JavaScript inside (fileList.hbs) is rather hacky. (i.e. no tests or so).

The tests for the generation and upload-code must be run manually. I think this is better than before (where there were no tests at all), but maybe we should use "mocha" or so. I don't think we should run these tests on every build.

With the deployed "index.json" file, we can integrate the file-list directly into the docs.

Please have a look and say what you think.

tasks/aws-s3-builds-page/fileList.hbs
```diff
@@ -0,0 +1,127 @@
+<!doctype html>
```
jaylinski, line 1: Maybe activate prettier formatting for this file?

package-lock.json
jaylinski, line 1: Did you update all deps? Seems to have lots of changes...

jaylinski (approved):
LGTM!

Issue #1991, opened 2023-09-02
Partials are too slow


mohd-akram:
While investigating why rendering a few hundred objects was incredibly slow in Handlebars, I found that the root cause is partials (profiling showed most of the time spent in `invokePartialWrapper`, a lot of time is also spent in `extend`). See the following [code](https://jsfiddle.net/9tfsoey1/):

```javascript
import Handlebars from 'handlebars';

const count = 10000;

// Handlebars
Handlebars.registerPartial('partial', Handlebars.compile('1234'));
const template = Handlebars.compile('{{> partial}}');

// Native
function partial() { return '1234'; }
const template2 = () => `${partial()}`;

// Warm up
for (let i = 0; i < count; i++) {
  template();
}

console.time('partial');
for (let i = 0; i < count; i++) {
  template();
}
console.timeEnd('partial');

// Warm up
for (let i = 0; i < count; i++) {
  template2();
}

console.time('native');
for (let i = 0; i < count; i++) {
  template2();
}
console.timeEnd('native');
```

Output:

```
partial: 50.114ms
native: 0.409ms
```

Partials are 120x times slower than a regular function call.

[Extended benchmark with indenting](https://jsfiddle.net/3sqcmrfj/)

mohd-akram:
Yes, I think this is as good as it's going to get for partials.

PR #1992, opened 2023-09-03
Improve partial indenting performance

mohd-akram:
This is about 30% faster in my testing.

Related to #1991.


jaylinski (approved):
Can confirm this is about 3 times as fast in Firefox.

mohd-akram:
Will probably have to revert this. Testing some more, this causes some excessive GC activity which actually makes things slower in some cases.

jaylinski:
Was reverted in #2042.

PR #1994, opened 2023-09-03
Reduce helper overhead

mohd-akram:
This results in about a 30% improvement in a helper-heavy template. `Utils.extend` is expensive to use in any hot path.


jaylinski (approved):
Tests pass, so I guess returning the `options`-object instead of a shallow clone also works.

kibertoad:
What is the status on this? Can we help to get this merged?

mohd-akram:
Moved to #1993.

PR #1995, opened 2023-09-03
Fix non-contiguous program indices

mohd-akram:


lib/handlebars/compiler/javascript-compiler.js
```diff
@@ -165,12 +165,10 @@ JavaScriptCompiler.prototype = {
 
       let { programs, decorators } = this.context;
       for (i = 0, l = programs.length; i < l; i++) {
-        if (programs[i]) {
```
jaylinski, line 168: Thinking out loud: since the `programs`-object isn't mutated inside the for-loop, I guess it's safe to omit this check.

lib/handlebars/compiler/javascript-compiler.js
```diff
@@ -914,8 +912,8 @@ JavaScriptCompiler.prototype = {
       let existing = this.matchExistingProgram(child);
 
       if (existing == null) {
-        this.context.programs.push(''); // Placeholder to prevent name conflicts for nested children
-        let index = this.context.programs.length;
+        // Placeholder to prevent name conflicts for nested children
+        let index = this.context.programs.push('') - 1;
```
jaylinski, line 916: `push()` returns the length, smart optimization!

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push#return_value

Thinking if we should keep the index at `1` and remove the `- 1`. :thinking:

jaylinski (approved):

lib/handlebars/compiler/javascript-compiler.js
mohd-akram: Yes, this check was due to the current implementation creating gaps in the array. The coverage tool complained, presumably because it always enters there so it's redundant, and I had to remove this to remain at 100% coverage.

PR #1996, opened 2023-09-04
Support `Map` and `Set` in #each block

jaylinski:
Support `Map` and `Set` in #each block and also support Map-keys in lookup-expressions.

See #1418 #1679

### DIFF
--- a/lib/handlebars/utils.js
+++ b/lib/handlebars/utils.js
@@ -35,14 +35,18 @@ export function isFunction(value) {
   return typeof value === 'function';
 }
 
-/* istanbul ignore next */
-export const isArray =
-  Array.isArray ||
-  function (value) {
+function testTag(name) {
+  const tag = '[object ' + name + ']';
+  return function (value) {
     return value && typeof value === 'object'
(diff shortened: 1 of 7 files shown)


lib/handlebars/utils.js
```diff
@@ -35,14 +35,18 @@ export function isFunction(value) {
...
     return value && typeof value === 'object'
-      ? toString.call(value) === '[object Array]'
+      ? toString.call(value) === tag
       : false;
   };
+}
+
+export const isArray = Array.isArray;
+export const isMap = testTag('Map');
+export const isSet = testTag('Set');
```
jaylinski, line 49: Memo: document new functions in docs and d.ts file.

https://handlebarsjs.com/api-reference/utilities.html#handlebars-utils-isarray-obj

jaylinski: See https://github.com/handlebars-lang/docs/pull/131/commits/b723f290eff868e3da03c19d39ab12d386c51ffd.

amanagr:
Waiting for this to get released!

PR #1998, opened 2023-09-07
Merge performance enhancements from `4.x`

jaylinski:
Cherry-picked changes from 4.x:
 
* #1995

### DIFF
--- a/lib/handlebars/compiler/javascript-compiler.js
+++ b/lib/handlebars/compiler/javascript-compiler.js
@@ -164,12 +164,10 @@ JavaScriptCompiler.prototype = {
 
       let { programs, decorators } = this.context;
       for (i = 0, l = programs.length; i < l; i++) {
-        if (programs[i]) {
-          ret[i] = programs[i];
-          if (decorators[i]) {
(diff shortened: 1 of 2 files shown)


kibertoad:
@jaylinski This looks cool! Any plans to finish this?

kibertoad:
@jaylinski do we have benchmarks for this?

jaylinski:
> @jaylinski do we have benchmarks for this?

Nope, I'm not even sure if the commit that is left does anything performance wise. I think it was just a cleanup to get the indices to start with `0`. So the plan is to merge it. Happy to hear your opinion.

kibertoad:
@jaylinski can we have a test for the change?

jaylinski:
@kibertoad What about the existing test that was adapted accordingly?

kibertoad:
@jaylinski which one is it?

jaylinski:
@kibertoad This one: `spec/expected/non.empty.amd.known.helper.js`

Apparently the change didn't affect any other tests.

kibertoad:
then probably we need new ones. amd is going away

PR #2004, opened 2023-09-19
Allow importing library as ESM

mohd-akram:
This allows importing the library directly from an ESM, without requiring a build step. One change was required for a test to pass in `index.runtime.js` - a copy of the runtime module object is made to allow overriding `checkRevision` (ESMs don't allow modifying import objects). The `source-map` module is also replaced in the build step rather than via runtime checking since `require` is not available in a ESM.

Before creating a pull-request, please check https://github.com/handlebars-lang/handlebars.js/blob/master/CONTRIBUTING.md first.

- [x] Have tests
- [x] Have the [typings](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html) (types/index.d.ts) updated on every API change. If you need help, updating those, please mention that in the PR description.


jaylinski:
Nice work, I had this on my backlog too. But we won't merge this in `4.x` in order to keep it backwards-compatible. Please target the PR at the `master`-branch.

mohd-akram:
Yes, I hoped it could be done in `4.x` but reached the same conclusion. Will be retargeting and adjusting the PR - it needs some changes most likely - but need to merge the CI and build PRs first.

mohd-akram:
Rebased on master, but closing for now as it depends on updating the build system, as well as fixing `@handlebars/parser`'s exports (it does not properly provide ESM support as it is missing an `exports` field in its package.json) for integration tests to pass.

PR #2005, opened 2023-09-19
Update babel and webpack

mohd-akram:
This is a prerequisite to #2004 and other potential enhancements.


mohd-akram:
Node.js 12, 14 and 16 should be dropped from the CI for pushes and pull requests against master since they're all EOL. They can be kept for pushes and pull requests against the 4.x branch only.

jaylinski:
The plan is to use https://github.com/developit/microbundle and get rid of grunt, babel and webpack.

If you wanna give it a try, please go ahead, otherwise please close it and wait for me to make the change.

mohd-akram:
Is it possible to merge this in the meantime until a switch to microbundle is made, since I already have it working with minimal changes and this is a blocker to other improvements. Also, looking at microbundle, it does not seem to be very regularly updated, and I'm not sure how flexible it is. If, for example, we want a bundle with some polyfills (I have a version of this PR for 4.x with all that, but it hit a snag with AMD support). I don't love grunt or babel, and would rip them out too, but that's a more complex endeavor at this point.

PR #2042, opened 2024-09-06
Revert "Improve partial indenting performance"

mohd-akram:
This reverts commit 08fddee0331cd52b82ced243dee2f0318600414f.

The change caused performance regressions in some cases and performance was not consistent across engines.


jaylinski:
Note to myself: also revert this in `master`. https://github.com/handlebars-lang/handlebars.js/pull/1998

PR #2137, opened 2026-03-19
Modernize S3 publishing

kibertoad:
* Remove grunt
* Add tests


package.json
```diff
@@ -77,6 +78,7 @@
     "concurrently": "^5.0.0",
     "eslint": "^10.0.3",
     "eslint-plugin-compat": "^7.0.1",
+    "fauxqs": "^2.3.1",
     "fs-extra": "^8.1.0",
```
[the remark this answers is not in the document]
kibertoad: we added conditional execution only on newer node versions for that exact purpose

tasks/publish-to-aws.js
```diff
@@ -4,6 +4,14 @@ const { S3, PutObjectCommand } = require('@aws-sdk/client-s3');
 const git = require('./util/git');
 const semver = require('semver');
 
+const BUCKET = 'builds.handlebarsjs.com';
+const PUBLISHED_FILES = [
+  'handlebars.js',
+  'handlebars.min.js',
+  'handlebars.runtime.js',
+  'handlebars.runtime.min.js',
+];
```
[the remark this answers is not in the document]
kibertoad: fixed

tasks/tests/publish-to-aws.test.js
```diff
@@ -0,0 +1,222 @@
...
+} = require('../publish-to-aws');
+
+let startFauxqs;
+let server;
+let s3Client;
+
+beforeAll(async () => {
+  ({ startFauxqs } = await import('fauxqs'));
+  server = await startFauxqs({ port: 0, logger: false });
+  server.createBucket(BUCKET);
```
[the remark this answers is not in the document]
kibertoad: fixed

kibertoad:
@jaylinski comments addressed, ready for rereview

package.json
```diff
@@ -77,6 +78,7 @@
     "concurrently": "^5.0.0",
     "eslint": "^10.0.3",
     "eslint-plugin-compat": "^7.0.1",
+    "fauxqs": "^2.3.1",
```
jaylinski: I'd prefer to not use this extra dependency, because it pulls in another 53 packages. And we already have way too much dev-dependencies.

I appreciate your effort to thoroughly test the AWS publishing script, but I'm not even sure if we will keep publishing to S3. So please remove the dependency and test for the AWS publishing script (or write a basic vitest test that mocks the AWS SDK).

(Nice package by the way! Noticed that you authored it.)

jaylinski (changes requested):

package.json
kibertoad: sure, I can replace it with some custom inline mock, as the surface that we need is pretty narrow

kibertoad: fixed

jaylinski (approved):

package.json
jaylinski: thanks!

PR #2139, opened 2026-03-22
chore: update type testing setup

mrazauskas:
Glad to see you found TSTyche useful. I am its author.

This PR updates and cleans up the type testing setup. I would suggest the following:

- removing `tstyche.config.json`, because TSTyche does all what is defined the by default;
- making `types/tsconfig.json` stricter;
- and removing `tsc` from the `lint:types` script, because `tstyche` does check the same files files as `tsc`.

### DIFF
--- a/types/tsconfig.json
+++ b/types/tsconfig.json
@@ -2,16 +2,14 @@
   "compilerOptions": {
     "module": "commonjs",
     "lib": ["es6"],
-    "noImplicitAny": true,
-    "noImplicitThis": true,
-    "strictNullChecks": true,
-    "strictFunctionTypes": true,
+    "exactOptionalPropertyTypes": true,
+    "noUncheckedIndexedAccess": true,
+    "strict": true,
     "noEmit": true,
 
-    "baseUrl": ".",
     "types": [],
     "paths": {
-      "handlebars": ["."]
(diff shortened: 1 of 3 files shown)


types/tsconfig.json
```diff
@@ -2,16 +2,14 @@
...
-    "noImplicitAny": true,
-    "noImplicitThis": true,
-    "strictNullChecks": true,
-    "strictFunctionTypes": true,
+    "exactOptionalPropertyTypes": true,
+    "noUncheckedIndexedAccess": true,
+    "strict": true,
     "noEmit": true,
 
-    "baseUrl": ".",
```
mrazauskas, line 11: `baseUrl` is deprecated in TypeScript 6 ([reference](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0-rc/#deprecated:---baseurl)).

types/tsconfig.json
```diff
@@ -2,16 +2,14 @@
...
     "types": [],
     "paths": {
-      "handlebars": ["."]
+      "handlebars": ["../"]
```
mrazauskas, line 12: Perhaps root of the repo instead of `./types` directory?

jaylinski:
Thank you!

PR #2148, opened 2026-04-12
fix: prevent proto-access bypass via Map Symbol.toStringTag spoofing

### LINKED ISSUE #2146: [Security] Proto-access control bypass via Map Symbol.toStringTag spoofing + HTML escape bypass
## Summary
Three security findings in Handlebars.js v5.0.0-alpha.1:

### Finding 1: Proto-Access Control Bypass via Fake Map (Medium)
**File:** `lib/handlebars/runtime.js:119-122`

`lookupProperty` completely skips proto-access controls for Map objects. Map detection uses `Object.prototype.toString` which is spoofable via `Symbol.toStringTag`:

```js
const fakeMap = { [Symbol.toStringTag]: 'Map', get(key) { return key === 'constructor' ? Function : undefined; } };
// lookupProperty(fakeMap, 'constructor') returns Function, bypassing the blocklist
```

**Fix:** Use `instanceof Map` instead of toString tag check.

### Finding 2: HTML Escaping Bypass via toHTML Duck-Typing (Medium)
**File:** `lib/handlebars/utils.js:61-66`

`escapeExpression` treats any object with a `toHTML` property as SafeString (no `instanceof` check). Attacker-controlled data with `toHTML` bypasses HTML escaping in `{{}}` expressions.

**Fix:** Use `instanceof SafeString` check.

### Finding 3: AST Injection via Direct AST Input (Medium)
**File:** `lib/handlebars/compiler/compiler.js:476-485`

`compile()` accepts pre-parsed AST objects without structural validation, enabling code injection via crafted AST nodes through the `Function()` constructor.

**Fix:** Validate AST structure before compilation.

eddieran:
Fixes the Map/Set tag-check bypass in `lookupProperty` where a user-controlled object can spoof `Symbol.toStringTag` to evade the proto-access guard.

(The `toHTML` / `SafeString` portion of the original PR was reverted in 0e4b21 — that change broke the documented duck-typing contract on `SafeString`. Will track that separately if needed.)

### What's in the diff
- `lib/handlebars/utils.js`: tag-check uses `Object.prototype.toString.call` only as a coarse guard, then verifies via `obj instanceof Map` / `instanceof Set` so a spoofed `@@toStringTag` on a plain object can't claim Map-ness.

Originally reported in handlebars-lang/handlebars.js#2146 (Finding 1).

### DIFF
--- a/lib/handlebars/utils.js
+++ b/lib/handlebars/utils.js
@@ -1,5 +1,3 @@
-import SafeString from './safe-string';
-
 const escape = {
   '&': '&amp;',
   '<': '&lt;',
@@ -37,15 +35,6 @@ export function isFunction(value) {
   return typeof value === 'function';
 }
 
-function testTag(name) {
-  const tag = '[object ' + name + ']';
-  return function (value) {
-    return value && typeof value === 'object'
-      ? toString.call(value) === tag
-      : false;
-  };
-}
-
 export const isArray = Array.isArray;
 export function isMap(value) {
   return value instanceof Map;
@@ -67,7 +56,7 @@ export function indexOf(array, value) {
 export function escapeExpression(string) {
   if (typeof string !== 'string') {
     // don't escape SafeStrings, since they're already safe
-    if (string instanceof SafeString) {
+    if (string && string.toHTML) {
       return string.toHTML();
     } else if (string == null) {
       return '';
--- a/lib/handlebars/utils.js
+++ b/lib/handlebars/utils.js
@@ -1,3 +1,5 @@
+import SafeString from './safe-string';
+
 const escape = {
   '&': '&amp;',
   '<': '&lt;',
@@ -45,8 +47,12 @@ function testTag(name) {
 }
 
 export const isArray = Array.isArray;
-export const isMap = testTag('Map');
-export const isSet = testTag('Set');
+export function isMap(value) {
+  return value instanceof Map;
+}
+export function isSet(value) {
+  return value instanceof Set;
+}
 
 // Older IE versions do not directly support indexOf so we must implement our own, sadly.
 export function indexOf(array, value) {
@@ -61,7 +67,7 @@ export function indexOf(array, value) {
 export function escapeExpression(string) {
   if (typeof string !== 'string') {
     // don't escape SafeStrings, since they're already safe
-    if (string && string.toHTML) {
+    if (string instanceof SafeString) {
       return string.toHTML();
     } else if (string == null) {
       return '';
--- a/spec/security.js
+++ b/spec/security.js
@@ -1,4 +1,24 @@
 describe('security issues', function () {
+  describe('GH-2146: Symbol.toStringTag-spoofed Map/Set should not bypass proto-access guards', function () {
+    it('should not treat a plain object spoofing Map via Symbol.toStringTag as a Map', function () {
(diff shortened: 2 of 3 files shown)


eddieran:
Scope-reduced per @copilot-pull-request-reviewer's note that the `instanceof SafeString` change was a breaking change to the documented `.toHTML` duck-typing contract. Reverted that piece; the remaining change is just the Map/Set Symbol.toStringTag spoofing fix (issue #2146 first bullet), which has no API impact. CI should be green now.

kibertoad:
The PR title and description still claim two fixes, but commit 0e4b21 reverted the toHTML → SafeString portion. 

The final diff is only the Map/Set tag-check fix in lib/handlebars/utils.js. Before merge, the title and body should be updated to match.

lib/handlebars/utils.js
```diff
@@ -35,18 +35,13 @@ export function isFunction(value) {
...
-export const isMap = testTag('Map');
-export const isSet = testTag('Set');
+export function isMap(value) {
+  return value instanceof Map;
+}
+export function isSet(value) {
+  return value instanceof Set;
+}
 
 // Older IE versions do not directly support indexOf so we must implement our own, sadly.
```
kibertoad, line 46: this needs a regression test, something like

```js
it('does not treat Symbol.toStringTag-spoofed objects as Map', function () {
    expectTemplate('{{value.constructor}}')
      .withInput({ value: { [Symbol.toStringTag]: 'Map', get: () => 'pwned' } })
      .toCompileTo(''); // or whatever the proto-controlled path produces
  });
```

kibertoad: Add one each for isMap / isSet at the unit level (spec/utils.js already has the happy-path tests) plus an end-to-end runtime.js lookup test.

eddieran:
Fair point — updated the title and body to match the actual diff (just the Map/Set tag-check fix). The `toHTML` piece is fully out of scope here.

eddieran:
Added the regression tests in 524d73d:

- `spec/utils.js` — `isMap` / `isSet` returning `false` for objects with `Symbol.toStringTag = 'Map'` / `'Set'`. The `Object.prototype.toString.call()` assertion in each is there to make the test fail loudly if anyone re-introduces the tag-based check.
- `spec/security.js` (new `GH-2146` block) — end-to-end: `{{constructor.name}}` and `{{lookup this "constructor"}}` resolve to `''` against a spoofed input, so the runtime proto-access guard is exercised through to the template output.

Local run: 83/83 pass on the node project.

Kelleretoro (approved):

Kelleretoro (approved):

eddieran:
Quick status update since the branch has been quiet: the PR title/body now match the reduced scope, and 524d73d adds both utility-level and end-to-end regression tests for the `Symbol.toStringTag` spoofing case.

Local node project tests pass (83/83). Happy to adjust further if maintainers want a different test shape.

kibertoad:
@eddieran I've thought about this one some more, and I have my doubts with regards to whether this needs to be addressed. Do I understand correctly that the only possible attack surface here is if attacker actually is capable to modify the source code of the application itself? If that happens, I don't think any kind of defensive programming would save us.

Can you describe a realistic scenario in which this change would actually be helpful?

PR #2150, opened 2026-04-15
Fix strict lookup in compat mode

### LINKED ISSUE #1741: Exception when the strict and compat compiler options are used together
Using `compat : true` and `strict : true` as compiler options causes and unexpected exception when executing the template

`require("handlebars").compile("{{v}}", { compat : true, strict : true })({ v : "a" })`

```
node_modules\handlebars\dist\cjs\handlebars\runtime.js:111
      if (!obj || !(name in obj)) {
                         ^

TypeError: Cannot use 'in' operator to search for 'v' in a
    at Object.strict (node_modules\handlebars\dist\cjs\handlebars\runtime.js:111:26)
    at Object.eval [as main] (eval at createFunctionContext (node_modules\handlebars\dist\cjs\handlebars\compiler\javascript-compiler.js:262:23), <anonymous>:11:87)
    at main (node_modules\handlebars\dist\cjs\handlebars\runtime.js:208:32)
    at ret (node_modules\handlebars\dist\cjs\handlebars\runtime.js:212:12)
    at ret (node_modules\handlebars\dist\cjs\handlebars\compiler\compiler.js:519:21)
    at Object.<anonymous> (index.js:9:80)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
```

Tested with version 4.7.7

https://runkit.com/embed/70ypjaj1dra0

### LINKED ISSUE #2149: Strict lookup doesn't work correctly in compat mode
Consider the template `{{#each items}}{{name}}{{/each}}`, executed with data like:

```js
{
  items: [
    { name: "Hello" }
  ]
}
```

When either the `strict` or the `compat` compile option is enabled on its own, the output is `"Hello"` as expected. But when _both_ `strict` and `compat` are enabled, the result is `TypeError: cannot use 'in' operator to search for "name" in "Hello"`.

Another example template where this issue occurs is `{{#foo}}Hello {{bar}}{{/foo}}`, executed with the data:
```js
{ foo: true, bar: "World" }
```

With either `strict` or `compat` enabled individually, the output is correctly `"Hello World"`. But when both `strict` and `compat` are enabled, the result is `TypeError: right-hand side of 'in' should be an object, got boolean`.

I noticed this since I am the maintainer of [PHP Handlebars](https://github.com/devtheorem/php-handlebars) and am working on implementing the `compat` option with the same semantics as in Handlebars.js, and ran into unexpected errors in Handlebars.js when comparing various templates compiled with both `compat` and `strict` enabled.

theodorejb:
Fixes #2149, fixes #1741

### DIFF
--- a/lib/handlebars/compiler/javascript-compiler.js
+++ b/lib/handlebars/compiler/javascript-compiler.js
@@ -1175,20 +1175,34 @@ function strictLookup(requireTerminal, compiler, parts, startPartIndex, type) {
     stack = compiler.nameLookup(stack, parts[i], type);
   }
 
-  if (requireTerminal) {
+  if (!requireTerminal) {
+    return stack;
+  }
+
+  if (startPartIndex > len) {
+    // Compat mode already consumed all parts via depthedLookup; the stack
+    // holds the resolved value, not an object to look the property up on.
+    // Use container.strictLookup to traverse depths with a strict error on miss.
     return [
-      compiler.aliasable('container.strict'),
-      '(',
-      stack,
-      ', ',
+      compiler.aliasable('container.strictLookup'),
+      '(depths, ',
       compiler.quotedString(parts[len]),
       ', ',
       JSON.stringify(compiler.source.currentLocation),
       ' )',
     ];
-  } else {
-    return stack;
   }
+
+  return [
+    compiler.aliasable('container.strict'),
+    '(',
+    stack,
+    ', ',
+    compiler.quotedString(parts[len]),
+    ', ',
+    JSON.stringify(compiler.source.currentLocation),
+    ' )',
+  ];
 }
 
 export default JavaScriptCompiler;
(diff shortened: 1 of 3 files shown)


kibertoad:
One suggestion: add a direct "{{v}} with strict+compat where v doesn't exist at any depth" test. Your "still throw when a property is missing at all depths" covers the same code path but via {{#each items}}, and a minimal top-level case would more clearly regress-guard the originally reported form.

theodorejb:
@kibertoad I rebased now and added a test case for a simple `{{v}}` template where the variable is missing.

kibertoad (approved):

kibertoad:
thank you!

PR #2151, opened 2026-04-30
Fix explicit null value handling in compat mode

theodorejb:
Follow-up to GH-2150.

`container.lookup` used `result != null` which caused explicit null values to fall through to parent contexts. This was inconsistent with expected behavior and with Mustache, which stops and returns a key when present in the object, regardless of its value. The whole point of `compat` mode is to match Mustache's lookup semantics, so I don't think this behavior was intended.

There were no tests for `compat` with an explicit null value, apart from a couple `compat` + `strict` mode tests I just added myself in GH-2150. One of these tests ("should still perform recursive lookup with a multi-part path not in context") was inconsistent with the test immediately below it ("should directly return an explicitly null property"), and that is fixed here.

This change makes it possible for unify the lookup logic in `container.lookup` and `container.strictLookup`, additionally fixing an issue where `strictLookup` didn't respect the `allowedProtoMethods` runtime option the same as `lookup` (see the added test in `spec/security.js` compared to the test above it).

Also fixed compat mode `Map` value resolution, and included the test for it from #2156 (if that is merged first, I can rebase this to fix the merge conflict).

### DIFF
--- a/lib/handlebars/runtime.js
+++ b/lib/handlebars/runtime.js
@@ -117,20 +117,10 @@ export function template(templateSpec, env) {
       return container.lookupProperty(obj, name);
     },
     strictLookup: function (depths, name, loc) {
-      const len = depths.length;
-      let depth;
-      for (let i = 0; i < len; i++) {
-        const d = depths[i];
-        if (
-          d &&
-          (typeof d === 'object' || typeof d === 'function') &&
-          name in d
-        ) {
-          depth = d;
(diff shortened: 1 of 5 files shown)


theodorejb:
@jaylinski @kibertoad By any chance can this be reviewed?

kibertoad (approved):

kibertoad:
thank you!

PR #2156, opened 2026-05-16
Fix compat Map depthed lookup

TorMatzAndren:
This fixes Map value resolution in compat-mode depthed lookup.

`container.lookup()` already resolves values through `container.lookupProperty()`, which supports `Map` entries. However, after finding a result it returned `depths[i][name]`, which reads from the object property instead of returning the resolved value.

That caused compat mode to fail for Map entries and behave inconsistently with normal Map lookup.

This change returns the already-resolved value and adds regression coverage.

Verified with:

- `npx vitest run --project node spec/basic.js spec/security.js spec/builtins.js`

### DIFF
--- a/lib/handlebars/runtime.js
+++ b/lib/handlebars/runtime.js
@@ -155,7 +155,7 @@ export function template(templateSpec, env) {
       for (let i = 0; i < len; i++) {
         let result = depths[i] && container.lookupProperty(depths[i], name);
         if (result != null) {
-          return depths[i][name];
+          return result;
         }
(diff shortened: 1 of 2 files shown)


theodorejb:
#2151 also fixes this, and I updated that PR to include your test. It might be better to merge this first, though, and then I can rebase my PR to keep it focused on fixing explicit `null` value handling in compat mode.

theodorejb:
This can be closed, since it was fixed as part of #2151.

PR #2173, opened 2026-07-22
docs: fix broken JavaScriptCompiler subclass example

### LINKED ISSUE #1912: The JavaScriptCompiler API docs sample code is broken as of 4.6
Just a note, that the [sample code for subclassing JavaScriptCompiler in the docs](https://github.com/handlebars-lang/handlebars.js/blob/master/docs/compiler-api.md#javascript-compiler) is broken as of ~4.6 with:

    TypeError: name.toLowerCase is not a function

(It works with handlebars 4.5 but fails with 4.6)

You can test this by swapping the handlebars.js version used in the [linked jsfiddle](https://jsfiddle.net/9D88g/162/) with the latest version (4.7.7).

Hashim1999164:
## Summary

- Update the JavaScriptCompiler subclass example in `docs/compiler-api.md` so it works on Handlebars 4.6+.
- Generate `lookupProperty(...)` with a lowercased name instead of calling a registered helper from `nameLookup`.
- Use `Object.create` for the prototype and note why the old helper-based approach fails.

Fixes #1912

## Test plan

- [ ] Confirm the compiler-api docs example matches current Handlebars compiler internals
- [ ] Spot-check that the documented approach works on Handlebars 4.6+

### DIFF
--- a/docs/compiler-api.md
+++ b/docs/compiler-api.md
@@ -321,13 +321,13 @@ The `Handlebars.JavaScriptCompiler` object has a number of methods that may be c
 ### Example for the compiler api.
 
 This example makes context property lookups case-insensitive by lowercasing the
-name at compile time, so `{{Test}}` resolves `test`. This illustrates how
-compiler behavior can be changed.
+name at compile time, so `{{#each Test}}` / `{{Value}}` resolve `test` / `value`.
+This illustrates how compiler behavior can be changed.
 
 Note: calling a registered helper from `nameLookup` (for example via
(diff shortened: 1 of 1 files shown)


Hashim1999164:
Updated the compiler example to lowercase the context name and then call the base `nameLookup`, and changed the description to use `{{#each Test}}` / `{{Value}}` instead of `{{Test}}`.

Hashim1999164:
Addressed the latest Copilot notes.

The example now says the compiler should resolve to the context, and MyCompiler.prototype.constructor is set after Object.create so the subclass still points at MyCompiler.

